PR #16, opened 2017-01-19
travis: cache Cargo dependencies and build output

mgeisler:
I hope this will speed up the build.


mgeisler:
Yep, this brings the build and test time down from about 8 minutes to just 1 minute!

PR #30, opened 2017-02-02
appveyor: build and test using beta and nightly

mgeisler:
This creates a matrix build on AppVeyor where the `%TOOLCHAIN%` environment variable changes from build to build. That variable is used with `rustup` to pick the right toolchain.


mgeisler:
The AppVeyor cache got cleared on the first build of this branch, but subsequent builds seem to use a per environment cache, which is exactly what we need.

PR #40, opened 2017-05-01
fix hyphenation example in README.md

ari7:
Thanks for `textwrap`!

I've fixed hyhenation example code.


mgeisler:
I've now released version 0.5.0 which brings the documentation on https://docs.rs/textwrap/ in sync with the README you see here on GitHub. So I don't think this PR is relevant any longer.

PR #49, opened 2017-05-18
NoHyphenation: non-splitting implementation of WordSplitter

mgeisler:
This type can be used in case words shouldn't be split at all, not
even on existing hyphens.


mgeisler:
Using this word splitter is analogous to `break_on_hyphens=False` in Python: https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.break_on_hyphens

PR #69, opened 2017-07-20
Rewrite wrap to borrow from input string

mgeisler:
This changes the implementation of `Wrapper::wrap` to borrow from the input string (where possible). The return type is now `Vec<Cow<str>>` instead of `Vec<String>` and so this can be a breaking change (but doesn't have to be if the lines are used in a way compatible with `Cow<str>`, such as how the `layout` example works).

Furthermore, the `squeeze_whitespace` option has been dropped since it seemed cumbersome to implement it with the new algorithm. If this is a concern, please let us know so we can supply a suitable work around.

### DIFF
--- a/benches/linear.rs
+++ b/benches/linear.rs
@@ -51,6 +51,32 @@ fn fill_800(b: &mut Bencher) {
     b.iter(|| textwrap::fill(text, LINE_LENGTH))
 }
 
+
+#[bench]
+fn wrap_100(b: &mut Bencher) {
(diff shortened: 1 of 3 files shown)


mgeisler:
The AppVeyor failures are due to a [problem with the AppVeyor firewall][1]. Since the TravisCI builds work as expected, I'll merge this PR.

[1]: http://help.appveyor.com/discussions/problems/7115-curl-cant-download-files-through-ssl#comment_43044339

Issue #59, opened 2017-05-23
wrap could return an iterator


mgeisler:
The `wrap` method currently returns `Vec<String>` but it could return an iterator instead. This would make it possible to output a large help text line-by-line and do less allocations.

However, if `wrap` is mostly called through `fill`, this might not really improve anything since `fill` still needs to copy the data into a new string.

mgeisler:
Fixed by #79.

Lokathor:
There doesn't seem to be an iterator based portion of the crate any more :(

mgeisler:
Hey @Lokathor, you're right. The iterator conflicted with the full-paragraph wrapping I implemented a while ago: instead of looking at the text word-by-word, the algorithm looks at the entire paragraph and finds globally optimal break points for it.

My expectation is that most strings are fairly small, so do you have a use-case where it's too expensive to keep a full `Vec<Cow<str>>` in memory together with the input `&str`?

Lokathor:
Naw it's a terminal program for desktop usage (a text game), so it's not too expensive in terms of space or speed. I just always look for the iterator version of stuff first and was sad to see that this iterator had disappeared at some point in the versioning.

mgeisler:
> I just always look for the iterator version of stuff first and was sad to see that this iterator had disappeared at some point in the versioning.

Yeah, I think it would be nice to support iteration, but it conflicts with [`wrap_optimal_fit`](https://docs.rs/textwrap/latest/textwrap/wrap_algorithms/fn.wrap_optimal_fit.html), which produces all wrapped lines at the same time.

Or rather: I guess an iterator could be added, with the understanding that you will be iterating over lines which are already formed in memory. The iterator could make sense for [`wrap_first_fit`](https://docs.rs/textwrap/latest/textwrap/wrap_algorithms/fn.wrap_first_fit.html), since that function can produce output line-by-line.

At a high level, I would be more interested in implementing #224, which is about wrapping while keeping state. So you could imagine feeding in text piece by piece and then getting the wrapped lines back in smaller chunks as well.

PR #77, opened 2017-08-24
Wordsplitter specialization

mgeisler:
This PR changes the `Wrapper` struct so it becomes generic over the type of word splitter used. I think this gives a somewhat better API, but it has the cost that one cannot change the word splitter used on an existing wrapper. I don't think that is something one would do very open, though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ use textwrap::Wrapper;
 
 fn main() {
     let corpus = hyphenation::load(Language::English_US).unwrap();
-    let wrapper = Wrapper::new(18).word_splitter(Box::new(corpus));
+    let wrapper = Wrapper::with_splitter(18, corpus);
     let text = "textwrap: a small library for wrapping text.";
     println!("{}", wrapper.fill(text))
 }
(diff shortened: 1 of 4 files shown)


hcpl:
Looking forward to this getting merged!

I already implemented `Wrapper::wrap_iter()` method but got stuck with `wrap_iter()` function because of transferring boxed trait objects to the iterator while making the iterator outlive a `Wrapper` instance. Since this PR removes boxing altogether, I can then proceed smoothly.

mgeisler:
@hcpl Hi! Oh, that's great to hear :-)

I'm still learning Rust, so any tips about how I can improve the API further will be very helpful...

hcpl:
I guess this is less about API being nice and more about practical reasons: since we can just create a new instance when we need another word splitter, we can eliminate redundant boxing. Which coincidentally enables some nifty features like cloning and allows the compiler to do more optimizations. And all that for free!

mgeisler:
@hcpl Yeah, makes sense! I wasn't able to measure any real performance improvements, though. But being able to clone a `Wrapper` sounds like a nice feature :-)

You mentioned a `wrap_iter` method -- would that basically be #59?

hcpl:
Yes, and I'm adapting my current implementation to changes now

PR #83, opened 2017-09-02
Make wrap_iter return a borrowing iterator

mgeisler:
Before, we had three methods that would take a string and return an iterator with wrapped lines:

* `wrap_iter(&self, &str)`: cloned self
* `wrap_iter_borrow(&self, &str)`: borrowed self
* `into_wrap_iter(self, &str)`: consumed self

Here, the `wrap_iter_borrow` method is preferable to `wrap_iter` if the `Wrapper` outlives the iterator.

I think that will be the usual case, and that we can do without the method that cloned self -- if needed, the caller can get the functionality back by combining `clone` and `into_wrap_iter`:
```
wrapper.clone().into_wrap_iter(s)
```
We can then rename `wrap_iter_borrow` to just `wrap_iter` and thus make the method with the short name do "the right thing" by default.

The new method names are thus:

* `wrap_iter(&self, &str)`: borrows self
* `into_wrap_iter(self, &str)`: consumes self

This makes the names a little more aligned with the standard library where `Vec::iter(&self)` also borrows self while you iterate over the collection.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -743,12 +743,18 @@ pub fn wrap(s: &str, width: usize) -> Vec<Cow<str>> {
 /// ```
 ///
 /// This function creates a Wrapper on the fly with default settings.
+/// It then calls the [`into_wrap_iter`] method. Hence, the return
+/// value is an [`IntoWrapIter`], not a [`WrapIter`] as the function
+/// name would otherwise suggest.
+///
 /// If you need to set a language corpus for automatic hyphenation, or
 /// need to wrap many strings, then it is suggested to create Wrapper
 /// and call its [`wrap_iter`] or [`into_wrap_iter`] methods.
 ///
 /// [`wrap_iter`]: struct.Wrapper.html#method.wrap_iter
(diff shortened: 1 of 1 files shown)


mgeisler:
@hcpl Please let me know how this PR looks to you -- will is mess up a use-case for you? Thanks!

hcpl:
`wrapper.clone().into_wrap_iter(s)` is a good way to reduce API bloat, really cool idea.

The only non-obvious thing I can find about this API implementation is that `wrap_iter` function returns `IntoWrapIter`, not `WrapIter`, which can be somewhat confusing for users. I think "Notice: this function returns `IntoWrapIter`, not `WrapIter`, as someone can expect due to its name mirroring `Wrapper::wrap_iter` method." can help.

Otherwise, everything looks good to me. Thanks for your involvement!

mgeisler:
Awesome, great that you like it :-) I'll add a comment to the `wrap_iter` function, thanks for the suggestion.

PR #86, opened 2017-09-02
Wrapper::wrap_iter: better document complexities

mgeisler:
We now describe the time and memory complexities in terms of both the
wrapping width and in terms of the input string length. This better
explains that fully stepping through the iterator requires O(1) extra
memory and O(*n*) time.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -340,6 +340,14 @@ impl<'w, 'a: 'w, S: WordSplitter> Wrapper<'a, S> {
     /// The result is a string with newlines between each line. Use
     /// the `wrap` method if you need access to the individual lines.
     ///
+    /// # Complexities
+    ///
+    /// This method simply joins the lines produced by `wrap_iter`. As
+    /// such, it inherits the O(*n*) time and memory complexity where
+    /// *n* is the input string length.
+    ///
+    /// # Examples
+    ///
     /// ```
     /// use textwrap::Wrapper;
     ///
@@ -347,10 +355,6 @@ impl<'w, 'a: 'w, S: WordSplitter> Wrapper<'a, S> {
     /// assert_eq!(wrapper.fill("Memory safety without garbage collection."),
     ///            "Memory safety\nwithout garbage\ncollection.");
     /// ```
-    ///
-    /// This method simply joins the lines produced by `wrap_iter`. As
-    /// such, it inherits the O(*n*) time and memory complexity where
-    /// *n* is the input string length.
     pub fn fill(&self, s: &str) -> String {
         let mut result = String::new();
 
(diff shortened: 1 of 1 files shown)


hcpl:
It should be noted that extra memory is O(1) in ideal case, when no indentation is provided.

Also, I think explanations for `fill` and `wrap` should be fixed too. Right now both have:
```
As such, it inherits the O(*n*) overall time and memory complexity
where *n* is the input string length.
```

I propose to use for `fill`:
```
As such, it inherits the O(*n*) overall time complexity
where *n* is the input string length
while also adds the O(*n*) memory complexity to store the obtained `String`.
```
and for `wrap`:
```
As such, it inherits the O(*n*) overall time complexity
where *n* is the input string length
while also adds the O(*k*) memory complexity to store the vector of lines
where *k* is the length of the vector.
```

mgeisler:
@hcpl Excellent points! I'll adjust the texts.

mgeisler:
> It should be noted that extra memory is O(1) in ideal case, when no indentation is provided.

Actually, I think this is technically correct as written: I wanted to describe the complexity in terms of the input string length *n*. If extra indentation is provided, you're correct that more memory will be allocated per line, but it's still a constant amount of memory (assuming a constant line width).

I've updated the text a little to maybe better reflect this.

mgeisler:
Heh, I looked a second time and you're right :-D So I've now revised the text to:

> This method returns a `WrapIter` iterator which borrows this
> `Wrapper`. Getting the next line from the iterator takes
> O(*w*) time and uses O(1) extra memory, where *w* is the
> wrapping width. This is in the ideal case with no extra
> indentation and no hyphenation, which means that the lines
> yielded are slices of the input string. If indentation is
> requested, O(*w*) extra memory will be used to store each
> line.
>
> In terms of the input string length *n* and with a constant
> wrapping width, the memory complexity is O(1) and the time
> complexity is O(*n*) when the iterator is fully processed.

mgeisler:
> while also (at least) adds the O(*k*) memory complexity to store the vector of lines
where *k* is the length of the vector.

I think this can be left as it is -- the *k* lines will each use about O(*n/k*) memory, so the whole vector will use about O(*n*) memory.

hcpl:
Vector of slices only uses O(*k*) memory though, since it only stores references to original strings.

mgeisler:
Oh... I hadn't thought the slices here! :-) Thanks again... I'll revise it again and then make a release tomorrow.

hcpl:
Well, maybe I'm too inclined into excessive details now. I just thought that finer-grained bound descriptions can be useful for those who want to grasp the inner workings quickly. But if that makes documentation itself more difficult, then I don't think these changes are worth it.

mgeisler:
> But if that makes documentation itself more difficult, then I don't think these changes are worth it.

Yeah, I agree! :-D I'll see if I can code up with a formulation that highlights the important point (linear running time).

PR #88, opened 2017-09-04
Release 0.8.0

mgeisler:

PR #89, opened 2017-09-04
Release 0.8.0

mgeisler:


mgeisler:
I forgot to update the version numbers in the README in #88, so this is the second attempt at the release :-)

Later, I saw that I had also forgotten to add the new links for the two issues fixed in version 0.8.0... That'll be fixed in the next release.

PR #90, opened 2017-09-09
Added regression test for case where width is set to usize::MAX

Fraser999:
As per [your comment](https://github.com/kbknapp/clap-rs/pull/1040#issuecomment-328258808), here's a PR with the same changes rebased on your current master branch.


mgeisler:
Thanks @Fraser999!

PR #95, opened 2017-09-18
lib: deny missing_debug_implementations

mgeisler:
This lint will ensure that all public types implement Debug, which in
turn makes them much more user friendly since they can be formatted
using {:?}.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,8 @@ This section lists the largest changes per release.
 
 ### Unreleased
 
+All public structs now implement `Debug`.
+
 Added a regression test for case where width is set to usize::MAX.
 Thanks @Fraser999!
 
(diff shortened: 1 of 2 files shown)


hcpl:
`Debug` impl for `CharIndices` appeared in [Rust 1.9](https://blog.rust-lang.org/2016/05/26/Rust-1.9.html) (relevant PR: https://github.com/rust-lang/rust/pull/32054, specifically for `CharIndices` [here](https://github.com/rust-lang/rust/pull/32054/files#diff-667f8101ac1fe1b2f63b7c2ef0691502R471)).

mgeisler:
Thanks @hcpl -- perhaps it's time to bump the oldest supported Rust version then...

mgeisler:
I ended up writing small `Display` implementations that ignore the `CharIndices`. I think that'll be good enough.

hcpl:
I'm sure this can be done better (e.g. for pretty-printing purposes) using [`DebugStruct`](https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html) - coming up with a PR right now!

PR #100, opened 2017-09-30
lib: implement Debug more elegantly

hcpl:
This way allows to pretty-print values (i.e. through `{:#?}` format specifier), if requested.


hcpl:
Err, something's up with `winapi`. Maybe caused by failed cache uncompression earlier.

mgeisler:
Ohh, this is much more clever than my approach :-D Thanks!

I've deleted the build cache for AppVeyor... it seems to become messed up from time to time.

PR #102, opened 2017-10-05
cargo: make term_size optional

### LINKED ISSUE #101: Remove `term_size` as a (hard required) dependency
Currently, `term_size` is a required dependency of this crate. This means that all consumers of this library **must** also install the transitive closure of `term_size`'s dependencies, even those who are not willing to use its functionality.

This is an especially important consideration given that `term_size` makes heavy use of `unsafe`, so crates with a zero-unsafe policy (such as one I am currently authoring) may not be able to use `textwrap` even though they don't actually need any of its `unsafe`-dependent functionality.

Since `with_termwidth()` is more or less trivial, and the `termwidth()` function literally just wraps `term_size` with a hardcoded default value, it seems to me that it is not the right trade-off that has been made in this regard.

Consequently, I suggest one of the following, in decreasing order of preference:
* Remove the `term_size` dependency, `termwidth()`, and `with_termwidth()` altogether, because the latter two are trivial and the default value of 80 might not be suitable for everyone anyway.
* Make `with_termwidth()` and `termwidth()` a non-default, optional feature and `term_size` an optional dependency, so they are not included by default, but those who are willing to use the functionality can still continue doing so without breaking code changes.
* Make `with_termwidth()` and `termwidth()` a default, optional feature and `term_size` an optional dependency, so they are included by default, but those who are not willing to use the functionality can opt out.

This is blocking https://github.com/kbknapp/clap-rs/issues/1055.

mgeisler:
The term_size crate can be used to determine the current terminal
width and we included a trivial convenience wrapper around it.

However, term_size makes heavy use of unsafe Rust and this was a
concern to some users of textwrap. This commit makes the dependency a
non-default optional feature.

Fixes #101.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ appveyor = { repository = "mgeisler/textwrap" }
 
 [dependencies]
 unicode-width = "0.1.3"
-term_size = "0.3.0"
+term_size = { version = "0.3.0", optional = true }
 hyphenation = { version = "0.6.1", optional = true }
 
 [dev-dependencies]
--- a/README.md
+++ b/README.md
@@ -30,6 +30,14 @@ dependency as:
 textwrap = { version = "0.8", features = ["hyphenation"] }
 ```
 
+To conveniently wrap text at the current terminal width, enable the
+`term_size` feature:
+
+```toml
+[dependencies]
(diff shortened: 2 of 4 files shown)


H2CO3:
@mgeisler Awesome, thanks a lot!

PR #120, opened 2018-01-01
bump rand to 0.4

ignatenkobrain:


mgeisler:
There were some problems with lack of disk space on Travis, so I've restarted the build. However, the failure on Rust 1.9.0 is genuine — any good reason not to support that old version of Rust? :-)

hcpl:
@mgeisler, likely to ease packaging `textwrap` -- @ignatenkobrain is a packager in Fedora community.

And since the clash of interests already happened I'm interested to learn about Fedora policies regarding old Rust versions as well.

ignatenkobrain:
In Fedora (and EPEL) we stick to latest versions of Rust ;)

mgeisler:
One of the dependencies bumped their minimum supported Rust version to 1.17.0, so I'll do the same... with that, the PR should become green :-)

mgeisler:
Thanks for the PR, I'm merging a rebased version here: #125.

Issue #60, opened 2017-05-23
Perhaps add a refill method


mgeisler:
It could be nice if textwrap supported "refilling" a paragraph. It should take an existing string like this
```rust
let help = "This is a long help text.
            It continues on this line
            since nicely-wrapped
            lines look better in the
            source code.

            This is a second paragraph
            in the same string.";
```

This string contains newlines and extra indentation. Refilling it would mean ignoring the extra newlines and indentation, and then otherwise wrapping it at some width.

Multiple paragraphs should probably be supported in order to make it extra convenient.

PR #123, opened 2018-01-18
WrapIterImpl: add support for inputs with line breaks

### LINKED ISSUE #122: Add support for line-breaks in 'fill' (support multi-line input)
```rust
println!("{}",fill("1 3 5 7
1 3 5 7",11));

/*
Expected:
1 3 5 7
1 3 5 7

Actual output:
1 3 5 7
1 3
5 7
*/
```

Currently, it seems that `fill` considers line breaks like any other whitespace character. It would be nice if the function could take line breaks into account when computing the wrapping!

axelchalon:
Closes #122 
does the trick but it needs reviewing!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -596,7 +596,7 @@ impl<'a> WrapIterImpl<'a> {
                 // we will return the line with its line break after exiting the loop
                 if self.split + self.split_len < self.source.len() {
                     let mut line = self.create_result_line(wrapper);
-                    cow_add_assign(&mut line, &self.source[self.start..self.split]);
+                    line += &self.source[self.start..self.split];
 
                     self.start = self.split + self.split_len;
                     self.line_width += wrapper.subsequent_indent.width();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -585,7 +585,27 @@ impl<'a> WrapIterImpl<'a> {
         while let Some((idx, ch)) = self.char_indices.next() {
             let char_width = ch.width().unwrap_or(0);
             let char_len = ch.len_utf8();
-            if is_whitespace(ch) {
+
+            if ch == '\n' {
+                self.split = idx;
+                self.split_len = char_len;
+                self.line_width_at_split = self.line_width;
+                self.in_whitespace = false;
+
+                // If this is not the final line, return the current line. Otherwise,
+                // we will return the line with its line break after exiting the loop
+                if self.split + self.split_len < self.source.len() {
+                    let mut line = self.create_result_line(wrapper);
+                    cow_add_assign(&mut line, &self.source[self.start..self.split]);
+
+                    self.start = self.split + self.split_len;
+                    self.line_width += wrapper.subsequent_indent.width();
+                    self.line_width -= self.line_width_at_split;
+
+                    return Some(line);
+                }
+            }
+            else if is_whitespace(ch) {
                 // Extend the previous split or create a new one.
                 if self.in_whitespace {
                     self.split_len += char_len;
@@ -1222,5 +1242,25 @@ mod tests {
         let y = vec!["\tfoo",
                      "  bar"];
         assert_eq!(dedent(&add_nl(&x)), add_nl(&y));
+    }
+
+    #[test]
+    fn multiline() {
+        assert_eq!(fill("1 3 5 7\n1 3 5 7",11),"1 3 5 7\n1 3 5 7");
+        assert_eq!(fill("1 3 5 7\n1 3 5 7",5),"1 3 5\n7\n1 3 5\n7");
+        assert_eq!(fill("1 3 5 7\n1 3 5 7",6),"1 3 5\n7\n1 3 5\n7");
+        assert_eq!(fill("1 3 5 7\nabcdefghi k",5),"1 3 5\n7\nabcde\nfghi\nk");
+        assert_eq!(fill("abcdefgh\ni k",5),"abcde\nfgh\ni k");
+        assert_eq!(fill("abcdefgh\nijklmnop",5),"abcde\nfgh\nijklm\nnop");
+
+        assert_eq!(fill("test\n",11),"test\n");
+        assert_eq!(fill("test\na",11),"test\na");
+        assert_eq!(fill("test\na\n",11),"test\na\n");
+        assert_eq!(fill("test\n\na\n\n",11),"test\n\na\n\n");
+        assert_eq!(fill("test\n\nabcdefghijk\n\nabcdefghijkm",11),"test\n\nabcdefghijk\n\nabcdefghijk\nm");
+        assert_eq!(fill("test\nabcdefghi\n",11),"test\nabcdefghi\n");
+        assert_eq!(fill("test\nabcdefghi\n\n",11),"test\nabcdefghi\n\n");
+        assert_eq!(fill("test\nabcdefghijklmnopq\n\n",11),"test\nabcdefghijk\nlmnopq\n\n");
+        assert_eq!(fill("test\nabcdefghijklmnopq abcdefghijk\n\n",11),"test\nabcdefghijk\nlmnopq\nabcdefghijk\n\n");
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -604,8 +604,7 @@ impl<'a> WrapIterImpl<'a> {
 
                     return Some(line);
                 }
-            }
-            else if is_whitespace(ch) {
+            } else if is_whitespace(ch) {
                 // Extend the previous split or create a new one.
                 if self.in_whitespace {
                     self.split_len += char_len;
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi Alex, thanks for the PR! I'm about to go to the US for 10 days so please don't be afraid if I don't get around to comment on their before February :-)

My first thought was that this is something the caller could handle by splitting the input string on newlines and then mapping `fill` over each line. It'll of course be less efficient, so maybe this is why you want it in line?

mgeisler:
By the way, the build failure seems to be caused by a breaking change in a dependency - not because of your changes. I'll have to sort that out...

axelchalon:
Hi Martin! Sure, no worries!

Yes, you're right that we could split the input first and then call `fill` on each line; however, I think it'd be more efficient if this was taken care of inside of `fill`.

Also, currently, calling `fill` on an input with line breaks yields unexpected results (i.e. output has line breaks at what seems at first like random positions, https://github.com/mgeisler/textwrap/issues/122), so even if we don't decide that `fill` preserve line breaks, then I imagine they should probably be at least stripped out so that we have a consistent result.

Have a nice trip!

mgeisler:
Hi Alex, I'm back again :-)

I think you're right: the current behavior is confusing... I had actually expected `fill` to convert `\n` into a space ` ` since the splitting is done on whitespace, not just on ` `. However, as you noticed, the `\n` is kept in the output and so weird line breaks show up!

I've considered adding a `refill` method (see #60) which would try to be "intelligent" and detect paragraphs in the input text. I would expect such a function to require a _blank line_ between paragraphs -- similarly to how Markdown separates paragraphs and how people have been formatting plaintext emails for decades. The reason: it would allow you to wrap the text nicely in your source file, while still have the text adapt to the user's terminal on display.

Do you think we should put that kind of logic into `fill` in all cases? That is, turn a single `\n` into a ` ` and turn `\n\n+` into what your patch treats like a single paragraph break?

Perhaps it's better to keep `fill` simple and just fix the annoying line breaks... a more fancy re-wrapping `refill` function could then be added on top. I would like a `refill` function to handle things like bullet lists too (more or less the same way Emacs handles these things out of the box).

src/lib.rs
```diff
@@ -1215,4 +1235,24 @@ mod tests {
                      "  bar"];
         assert_eq!(dedent(&add_nl(&x)), add_nl(&y));
     }
+
+    #[test]
+    fn multiline() {
+        assert_eq!(fill("1 3 5 7\n1 3 5 7",11),"1 3 5 7\n1 3 5 7");
```
mgeisler: Could you run `rustfmt` on the file? I believe it would insert a space after the `,`. Thanks!

src/lib.rs
```diff
@@ -601,7 +601,27 @@ impl<'a> WrapIterImpl<'a> {
...
+
+                // If this is not the final line, return the current line. Otherwise,
+                // we will return the line with its line break after exiting the loop
+                if self.split + self.split_len < self.source.len() {
+                    let mut line = self.create_result_line(wrapper);
+                    cow_add_assign(&mut line, &self.source[self.start..self.split]);
+
+                    self.start = self.split + self.split_len;
+                    self.line_width += wrapper.subsequent_indent.width();
+                    self.line_width -= self.line_width_at_split;
```
mgeisler: I believe this is redundant -- in total, `self.line_width` becomes `wrapper.subsequent_indent.width()` since you set it to `self.line_width_at_split` above. So you can remove this line and simplify the previous line to a direct assignment: 
```
self.line_width = wrapper.subsequent_indent.width();
```

src/lib.rs
```diff
@@ -1215,4 +1235,24 @@ mod tests {
                      "  bar"];
         assert_eq!(dedent(&add_nl(&x)), add_nl(&y));
     }
+
+    #[test]
+    fn multiline() {
+        assert_eq!(fill("1 3 5 7\n1 3 5 7",11),"1 3 5 7\n1 3 5 7");
+        assert_eq!(fill("1 3 5 7\n1 3 5 7",5),"1 3 5\n7\n1 3 5\n7");
+        assert_eq!(fill("1 3 5 7\n1 3 5 7",6),"1 3 5\n7\n1 3 5\n7");
```
mgeisler: It's a little hard for me to figure out which corner-case each test is testing :-) Could you perhaps add multiple smaller tests if there is a common theme?

Here, I think a third line with
```
        assert_eq!(fill("1 3 5 7\n1 3 5 7", 7), "1 3 5 7\n1 3 5 7"); 
```
could be interesting -- it would demonstrate that the code doesn't accidentally add an extra `\n`.

src/lib.rs
```diff
@@ -1215,4 +1235,24 @@ mod tests {
...
+
+        assert_eq!(fill("test\n",11),"test\n");
+        assert_eq!(fill("test\na",11),"test\na");
+        assert_eq!(fill("test\na\n",11),"test\na\n");
+        assert_eq!(fill("test\n\na\n\n",11),"test\n\na\n\n");
+        assert_eq!(fill("test\n\nabcdefghijk\n\nabcdefghijkm",11),"test\n\nabcdefghijk\n\nabcdefghijk\nm");
+        assert_eq!(fill("test\nabcdefghi\n",11),"test\nabcdefghi\n");
+        assert_eq!(fill("test\nabcdefghi\n\n",11),"test\nabcdefghi\n\n");
+        assert_eq!(fill("test\nabcdefghijklmnopq\n\n",11),"test\nabcdefghijk\nlmnopq\n\n");
+        assert_eq!(fill("test\nabcdefghijklmnopq abcdefghijk\n\n",11),"test\nabcdefghijk\nlmnopq\nabcdefghijk\n\n");
```
mgeisler: The length of `"abcdefghijklmnopq"` is 17, so I'm not actually sure if there is a relation between that number and the width of 11?

src/lib.rs
```diff
@@ -601,7 +601,27 @@ impl<'a> WrapIterImpl<'a> {
...
+
+            if ch == '\n' {
+                self.split = idx;
+                self.split_len = char_len;
+                self.line_width_at_split = self.line_width;
+                self.in_whitespace = false;
+
+                // If this is not the final line, return the current line. Otherwise,
+                // we will return the line with its line break after exiting the loop
+                if self.split + self.split_len < self.source.len() {
```
mgeisler: I didn't actually try, but would it be possible to combine the logic here with the existing logic that kicks in when the current line it too long? Maybe by changing the conditions slightly:
```
if ch != '\n' && is_whitespace(ch) {
    // ...
} else if ch == `\n` || self.line_width + char_width > wrapper.width {
    // ...
}
```

No idea if it works! :-D

axelchalon:
Thanks for the feedback! I'm a bit busy at the moment but I'll get back to you in a bit

mgeisler:
Hi @axelchalon, no stress at all! Let me know if you think my suggestions are annoying or unhelpful, in which case I'll be happy to do them myself. Otherwise just take your time to iterate on this.

I should have fixed the brokenness of `master` by bumping the minimum version of Rust we test against. If you rebase your branch and force push, you should see tests passing.

axelchalon:
Hey Martin! It's all good, no worries :) 

> I would expect such a function to require a blank line between paragraphs [...] The reason: it would allow you to wrap the text nicely in your source file, while still have the text adapt to the user's terminal on display.

Hmm, Rust already supports escaping line breaks and subsequent leading whitespace with `\` at the end of a line. Also, I would expect such issues to be handled by the language itself, and I feel that libraries shouldn't make perhaps less intuitive decisions to make the language more easy-to-use? I guess we can count on the rust dev team to keep working on this?
I understand where you're coming from, but I think the library should stick to its primary purpose; I see the library more as a util library for quick text reformatting / simple text manipulation, not so much as a pretty printer / library for markdown-ey parsing of human-formatted input. As such I would expect the library to execute minimal tasks and for its functions to do a single thing (and do it well! :D). I would expect the `fill` function to simply wrap the input (and keep the line breaks). For example if I want to wrap the following string (the help message of a CLI option):
```
Set the operating mode. MODE can be one of:
    last - Uses the last-used mode, active if none;
    active - Parity continuously syncs the chain;
    passive - Parity syncs initially, then sleeps and wakes regularly to resync;
    dark - Parity syncs only when the RPC is active;
    offline - Parity doesn't sync.",
```
It certainly wouldn't feel intuitive for me to have to separate each line with an extra blank line. I would expect the function to simply do what it's supposed to do: wrap the lines that are too long. (Of course, we could say that this function could handle lists as well, but then again it would overreach its simple purpose / intended function.)

src/lib.rs
axelchalon: hmm maybe but I can't really seem to get it working this way ^^ feel free to commit if you find a way

mgeisler:
> I understand where you're coming from, but I think the library should stick to its primary purpose; I see the library more as a util library for quick text reformatting / simple text manipulation, not so much as a pretty printer / library for markdown-ey parsing of human-formatted input.

Very good point! More elaborate wrapping logic should at the very least be put into a different function then.

mgeisler:
Thanks a lot, I've merged the branch and will make a release in one of the next days.

PR #125, opened 2018-02-13
cargo: bump rand dev-dependency to latest version

mgeisler:
Since we're now using Rust 1.17 or later, we can start using the
latest version of rand for our tests.


mgeisler:
Supersedes #120.

Issue #126, opened 2018-03-30
Allow the use of the library outside of the context of the terminal


Moxinilian:
Hello, this is a feature request.

It could be really useful to be able to use this library outside of the context of the CLI.
I'm thinking about Piston text, that provides rendered characters with their widths depending on font-size, TrueType Font, etc... I just browsed the source code, and I believe supporting it would only be a matter of allowing to insert custom sizes for characters, but I am not sure as I obviously do not grasp all of the big picture. It would also require to handle floating point numbers in some way, as maybe the size could not be an integer.
If you believe it's a good idea and is easy enough, I could do it myself.

Thanks in advance!

Moxinilian:
On the matter of the API, I would make Wrapper take an optional closure taking a character and returning its width. Do you think it is reasonable?
I was also thinking that computing the size of characters could potentially be expensive (especially for TrueType Font for example). Do you think it is a relevant issue?
If so, do you think caching the size of characters has its place in the library, or should it instead be handled by the closure?

Moxinilian:
So I made a PR to be able to provide custom width for characters, and I am now considering support for floating points. But considering the nature of the change, it would force it to be breaking and therefore force on people the disadvantages of floating points. I could try to support both but that would obviously have an impact on the simplicity of the code.
What do you think would be preferable?

mgeisler:
Hi @Moxinilian and @robinkrahl, as mentioned in #128, I would still like to see textwrap used outside the context of the terminal. So some way of measuring the displayed width of non-console text seems like a great first step.

As mentioned in https://github.com/mgeisler/textwrap/pull/128#issuecomment-711286080, I could imagine expanding the `WrapOptions` trait with a `width(&str) -> T` method, which would measure a single word, or [rather a grapheme](https://docs.rs/unicode-segmentation/), since I want to start using graphemes as the unit when breaking apart long words.

The implementation of `WrapOptions` for `usize` would then change to something along the lines of
```
impl WrapOptions<usize> for usize {
    fn width(&self, word: &str) -> usize {
        word.width()
    }
}
```
Writing this makes me realize that there is already a `width() -> usize` method on the `WrapOptions` trait :smile:  That's a bit silly, so we would need a better word for one of the two...

The `Options` struct could be parameterized by `T` and it could gain a `width_provider` method which takes a closure which measures the width. This would move us back towards the old API where `Wrapper<T>` was parameterized on the type of "word splitter" used.

[...]

PR #128, opened 2018-04-01
Added support for custom width providers

Moxinilian:
This is part of the ongoing effort to support #126
This PR offers a way to set custom (integer) widths for characters. It does not add support for floating point numbers yet.
This allows the user to pass to most functions a closure taking a ``char`` and returning its width as a ``usize``.
I tried to limit API changes for simple tasks by leaving the ``fill`` and ``wrap`` functions intact in signature, and added ``wrap_dynamic`` and ``fill_dynamic`` that take the closure.
I was not really inspired with those names, if you want me to change them to something else I would be glad to do it.
I also chose to re-export ``unicode_width::UnicodeWidthChar`` as it will probably come in handy for people writing custom closures, avoiding them to depend on ``unicode_width``.

Example:
```rust
let wrap = Wrapper::new(15);
let res = wrap.wrap_dynamic("Concurrency without data races.", |c| {
    if c == 'h' {
        c.width().unwrap_or(0) + 4
    } else {
        c.width().unwrap_or(0)
    }
});
assert_eq!(res, vec!["Concurrency", "without", "data races."]);
```
This example illustrates an hypothetical case where an h would actually take 5 columns instead of 1.
I included it as a test, with its equivalent for ``Wrapper::fill_dynamic``.

Benchmark:
```
before
test fill_100 ... bench:         539 ns/iter (+/- 115)
test fill_200 ... bench:       1,188 ns/iter (+/- 141)
test fill_400 ... bench:       2,416 ns/iter (+/- 329)
test fill_800 ... bench:       5,349 ns/iter (+/- 974)
test wrap_100 ... bench:         639 ns/iter (+/- 111)
test wrap_200 ... bench:       1,397 ns/iter (+/- 206)
test wrap_400 ... bench:       2,729 ns/iter (+/- 405)
test wrap_800 ... bench:       5,575 ns/iter (+/- 664)

after
test fill_100 ... bench:         554 ns/iter (+/- 139)
test fill_200 ... bench:       1,199 ns/iter (+/- 145)
test fill_400 ... bench:       2,409 ns/iter (+/- 1,570)
test fill_800 ... bench:       5,142 ns/iter (+/- 658)
test wrap_100 ... bench:         632 ns/iter (+/- 113)
test wrap_200 ... bench:       1,430 ns/iter (+/- 221)
test wrap_400 ... bench:       2,732 ns/iter (+/- 323)
test wrap_800 ... bench:       5,417 ns/iter (+/- 456)
```

Zero-cost abstraction really is awesome.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,7 +55,7 @@ use std::borrow::Cow;
 use std::str::CharIndices;
 
 use unicode_width::UnicodeWidthStr;
-use unicode_width::UnicodeWidthChar;
+pub use unicode_width::UnicodeWidthChar;
 #[cfg(feature = "hyphenation")]
 use hyphenation::{Corpus, Hyphenation};
 
@@ -224,6 +224,10 @@ pub struct Wrapper<'a, S: WordSplitter> {
     pub splitter: S,
 }
 
+fn default_width_provider(a: char) -> usize {
+    a.width().unwrap_or(0)
+}
+
 impl<'a> Wrapper<'a, HyphenSplitter> {
     /// Create a new Wrapper for wrapping at the specified width. By
     /// default, we allow words longer than `width` to be broken. A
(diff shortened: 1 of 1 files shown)


mgeisler:
Hey @Moxinilian, thanks for the PR. It's not  been forgotten, I've just been busy lately...

I'm traveling this weekend, so it won't be before next week that I get proper time to look at this.

Just some short remarks: I love that you could do this via a generic function parameter --- zero-cost abstractions are awesome, as you say :-)

About measuring the characters: would it perhaps be better to measure words at a time?

With combining characters, I guess you don't really know the true displayed width of a word until you typeset the entire thing. Take emoticons as an extreme example since I'm typing this on my phone: I recently learned that a symbol like 👩‍👩‍👧‍👦 is created at rendering time --- it's really multiple Unicode characters with zero-width-joiners between them, something like 👧👧👨👨 next to each other. See https://emojipedia.org/emoji-zwj-sequences/.

So if you measure the width of this `char` by `char`, then you get the wrong result. I think other combining characters will have the same problem.

Finally, you're definitely right that the crate should be refactored by now... It was my first larger piece of Rust code, so I'm happy to learn how to structure it better!

Moxinilian:
I'm going to revert the changes and try to implement that instead, as there would be no point in merging this now that we're going down the word route.

Moxinilian:
I'm going to read this document to verify easy solutions have not already been found.
http://unicode.org/reports/tr14/

mgeisler:
I'm thinking I'll make a 0.10.0 release now (after #129 got reported) and then we can merge this in and play a bit with the API and do refactorings and whatnot for a later 0.11.0 release. Hope that seems okay!

robinkrahl:
I’m also interested in this feature to be able to use `textwrap` with PDF documents (in [`genpdf`](https://lib.rs/crates/genpdf)).  What is the current status of this PR?  Is there anything I can do to help?

mgeisler:
Hi @robinkrahl, I think this PR has stalled, but I would still like to see this kind of feature implemented. @Moxinilian, if you're still interested in this, then please let us know!

I recently rewrote some of the inner parts of textwrap (#213), so I think this PR should be closed and then a new one could be opened to fit with the new API. Ideally, this change should be invisible at the surface-level of the API, so that one can continue to write

```
let lines = wrap("some text", 80);
```

Perhaps the `WrapOptions` trait could gain a `width(&str) -> T` method, where `T` can be `usize`, `float`, or even `u16` as needed (I mention `u16` since the `terminal_size` crate [measures the width in  `u16`](http://eminence.github.io/terminal-size/doc/terminal_size/struct.Width.html), which makes me think that it's completely overkill to use `usize` all over).

Let me close this PR for now and then we can perhaps discuss more in #126.

PR #141, opened 2018-07-04
Fix dedent handling of empty lines and nl

bbqsrc:
Hey!

I noticed a couple of issues with trailing newlines and blank lines with `dedent`, so I created a few new test cases and then made them pass.

Also added a couple of checks for mixed whitespace patterns to make sure I didn't break anything while I was at it. :smile:


mgeisler:
Hej @bbqsrc, thank you very much for the PR -- and sorry for the slow turn-around time.

The build failure is unrelated, it seems that some library has again bumped the minimum required rustc version. I'll likely have to do the same for textwrap then.

PR #149, opened 2018-09-19
Add function for bulleted lists and enumerations

jomala:
I think this is a useful addition. Are you happy that it fits?

### DIFF
--- a/examples/termwidth.rs
+++ b/examples/termwidth.rs
@@ -21,7 +21,10 @@ fn main() {
     #[cfg(feature = "hyphenation")]
     fn new_wrapper<'a>() -> (&'static str, Wrapper<'a, hyphenation::Corpus>) {
         let corpus = hyphenation::load(Language::English_US).unwrap();
-        ("with hyphenation", Wrapper::with_splitter(textwrap::termwidth(), corpus))
+        (
+            "with hyphenation",
+            Wrapper::with_splitter(textwrap::termwidth(), corpus),
+        )
     }
 
     let example = "Memory safety without garbage collection. \
--- a/src/indentation.rs
+++ b/src/indentation.rs
@@ -72,16 +72,16 @@ pub fn indent(s: &str, prefix: &str) -> String {
 /// ```              
 pub fn bullet(s: &str, prefix: &str) -> String {
     let mut result = String::new();
-	let prefix_len = UnicodeWidthStr::width(prefix);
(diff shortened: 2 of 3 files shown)


mgeisler:
Hi @jomala, thank you for the pull request.

The code looks nice. However, it is actually possible to use the `initial_indent` and `subsequent_indent` fields on `Wrapper` to achieve something similar: https://docs.rs/textwrap/0.10.0/textwrap/struct.Wrapper.html#examples-1

Do you think it makes sense to have a stand-alone `bullet` function then? I'm not sure, since I think people will often want to both _wrap_ and _format_ text into a bullet list. In some sense, the `indent` function is similarly redundant since you can achieve the same effect by wrapping with `initial_indent` and `subsequent_indent` set to a string with the desired number of spaces.

src/indentation.rs
```diff
@@ -59,6 +61,33 @@ pub fn indent(s: &str, prefix: &str) -> String {
...
+/// ```
+/// use textwrap::bullet;
+///
+/// assert_eq!(bullet("The first line\nand some text\ngiving more details.", "  - "),
+///            "  - The first line\n    and some text\n    giving more details.\n");
+/// ```              
+pub fn bullet(s: &str, prefix: &str) -> String {
+    let mut result = String::new();
+    let prefix_len = UnicodeWidthStr::width(prefix);
+    let spaces = format!("{:w$}", "", w = prefix_len);
```
mgeisler: I think this can be written easier as
```
let spaces = " ".repeat(prefix.width());
```
The `prefix_len` variable is then not needed (the `UnicodeWidthStr` trait adds a `width` method to `str`).

examples/termwidth.rs
```diff
@@ -21,7 +21,10 @@ fn main() {
     #[cfg(feature = "hyphenation")]
     fn new_wrapper<'a>() -> (&'static str, Wrapper<'a, hyphenation::Corpus>) {
         let corpus = hyphenation::load(Language::English_US).unwrap();
-        ("with hyphenation", Wrapper::with_splitter(textwrap::termwidth(), corpus))
+        (
```
mgeisler: Can you take out this change? It looks unrelated to what the PR does.

jomala:
I hadn't actually noticed those options on Wrapper. It might be worth highlighting them in the docs.

But apart from that point, I don't think this merge does have too much merit then. Let's just close it.

Thanks!

Issue #148, opened 2018-09-08
hyphenation is no longer maintained


BurntSushi:
It appears that [`hyphenation` is no longer being maintained](https://github.com/tapeinosyne/hyphenation/issues/12), and as a result, it is blocking ripgrep from being [packaged in Debian](https://github.com/BurntSushi/ripgrep/issues/10#issuecomment-419611000). I'm opening this issue to see if you're willing to brainstorm ideas on a path forward here.

Unless hyphenation becomes actively maintained, I think the only path forward is to find a way to remove it from `textwrap`'s dependency tree. Here are some ideas:

* `textwrap` could vendor `hyphenation`'s functionality and provide it itself. This means the maintenance burden gets shifted to the maintainers of this crate.
* Someone forks and maintains `hyphenation` by publishing a different crate that provides the same functionality, and then `textwrap` switches to that fork. This means the maintenance burden gets shifted to whoever makes the fork.
* I fork `hyphenation` and maintain it myself. I'd prefer not to do this, because if I do, I'm going to want to dig into it and solve the problem from first principles. (I don't know whether the existing implementation is the same one that I would arrive at.) However, it might be possible that `textwrap` only needs a small piece of functionality from `hyphenation`. If so, perhaps I could just provide that, but I haven't investigated yet.

What are your thoughts?

mgeisler:
Hi Andrew, thanks for bringing this up! And thanks for Ripgrep -- I use it every day :-D

The auto-hyphenation feature via the `hyphenation` craete is an optional dependency for `textwrap`, so I believe my crate could just be packaged without it. As for `ripgrep`, I suppose the dependency on `textwrap` is through `clap`? The `clap` create does not expose a way to configure or use the `hyphenation` crate, so any packaging should hopefully be able to take the `hyphenation` crate out.

It's a shame to see the crate fall behind, though! I think it's a super fun feature to have automatic hyphenation of things like help texts :-) I would be happy to work with @tapeinosyne and perhaps take over maintenance if that's the best way forward. I haven't looked super deep into the internals of that crate, though, but I would be interested in doing so and keep the crate up to date.

I'll go read https://github.com/BurntSushi/ripgrep/issues/10 now to see why `hyphenation` is deemed necessary to package.

mgeisler:
[...]

PR #150, opened 2018-09-24
Update `hyphenation` to 0.7.1

tapeinosyne:
Hyphenation dictionaries are now serialized with stable serde, thus addressing #148.


tapeinosyne:
Version v0.7.0 had an outdated downstream dependency on `num-traits`, as well as a minor compatibility issue with Rust 1.21.0.  Both issues have been addressed in v0.7.1.

mgeisler:
Wow, thanks for an awesome PR!

PR #168, opened 2019-08-04
cargo: update version-sync dep to latest version

mgeisler:


mgeisler:
This makes #159 redundant.

Issue #140, opened 2018-05-22
Handle ANSI color codes


mgeisler:
Textwrap currently doesn't know about ANSI color codes for colored terminal output. It simply treats these escape codes as any other character and this can mess up the output as shown here: kbknapp/clap-rs#1246.

The custom width functionality of #128 could perhaps be used to solve this problem, but I feel it would be nice if this was a builtin feature (perhaps an optional feature depending on how invasive it will be to handle this).

laanwj:
Yes! this would be very useful for me too, though in my case the codes are not *ANSI* color codes but other game-specific formatting.

A general way to mark parts of the input text as color/formatting would be nice.

Such spans would be ignored for purposes of width estimation, formatting and breaking, but would be emitted in the output in the same place as they were in the input.

laanwj:
Oh wow, hadn't thought of that: as long as the textwrap doesn't change the text, it doesn't have to know about the formatting at all, the string is divided in the same places!

It's unfortunate that this doesn't extend to hyphen-ization or multiple spaces—I guess even that could be fixed up after the fact with a reconciliation pass—though a solution that works on tokens, as you say, would be more elegant.

mgeisler:
I hadn't though of it before you gave me the idea :-D I think you could take the above idea pretty far, but it'll be much nicer if there was some builtin support. I'll see what I can come up with...

medusacle:
might be useful to take inspiration from how cursive handles this for their TextViews, as it more or less exactly implements the above;

they have a `StyledString = SpannedString<Style>` which consists of spans: https://github.com/gyscos/Cursive/blob/master/src/utils/markup/mod.rs#L20, these are parametrized on the struct to be used as style, cursive itself uses https://github.com/gyscos/Cursive/blob/master/src/theme/style.rs#L8
e.g.
```rust
    let mut styled = StyledString::plain("Isn't ");
    styled.append(StyledString::styled("that ", Color::Dark(BaseColor::Red)));
    styled.append(StyledString::styled(
        "cool?",
        Style::from(Color::Light(BaseColor::Blue)).combine(Effect::Bold),
    ));
```
then their word wrapper  which operates on them is here: https://github.com/gyscos/Cursive/blob/master/src/utils/lines/spans/lines_iterator.rs#L13

pksunkara:
@mgeisler Any update on this? Would like to get it fixed on clap.

mgeisler:
[...]

PR #179, opened 2020-04-15
lib: ignore ANSI escape sequences

mgeisler:
ANSI escape sequences are typically used for colored text. The
sequences start with a so-called CSI, followed by some "parameter
bytes" before ending with a "final byte".

We now handle these escape sequences by simply skipping over the
bytes. This works well for escape sequences that change colors since
they don't take up space and since they continue to work across any
line breaks we insert.

See https://en.wikipedia.org/wiki/ANSI_escape_code for details.

Fixes: #140.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -85,6 +85,13 @@ use unicode_width::UnicodeWidthStr;
 /// A non-breaking space.
 const NBSP: char = '\u{a0}';
 
+/// The CSI or "Control Sequence Introducer" introduces an ANSI escape
+/// sequence. This is typically used for colored text and will be
+/// ignored when computing the text width.
+const CSI: (char, char) = ('\u{1b}', '[');
+/// The final bytes of an ANSI escape sequence must be in this range.
+const ANSI_FINAL_BYTE: std::ops::RangeInclusive<char> = '\x40'..='\x7e';
+
 mod indentation;
 pub use crate::indentation::dedent;
 pub use crate::indentation::indent;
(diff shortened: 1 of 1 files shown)


mgeisler:
@pksunkara, would you be able to do a test with this commit and tell me if this solves your problem in  clap-rs/clap#1246?

mgeisler:
This PR will of course not do anything sensible for Windows. As far as I understand, colored text depends on using special Windows APIs and so I guess the color information isn't embedded into the string like on Linux?

pksunkara:
I can't say because we refactored our color writing and we don't construct strings anymore but write directly to `io`. So, all the strings we now give to textwrap are plain.

mgeisler:
Aha, I thought based on https://github.com/clap-rs/clap/issues/1246 and https://github.com/mgeisler/textwrap/issues/140#issuecomment-612460117 that `clap` was needing this feature? I was hoping you could test it for me to help guide it and make sure it's useful.

mgeisler:
I'll merge it right after I fix Travis -- my crate has again stopped building on Rust 1.31.0.

pksunkara:
Yeah, we will probably need it again when we are fixing https://github.com/clap-rs/clap/issues/1456. But, unfortunately, it is not a priority until after we release 3.0

mgeisler:
Okay, cool! I'll get a release out of textwrap in the meantime and then I hope it'll be useful for you guys.

pksunkara:
Before you release, do you think you can port this to https://lib.rs/crates/terminal_size? `term_size` is not being maintained and I would appreciate if you could port it to a more maintained version.

mgeisler:
Sure, let me update that right away.

PR #209, opened 2020-09-17
Make indent preserve the of \n in the input string (stop adding a trailing \n)

### LINKED ISSUE #207: The indent function always adds a final \n, even if the input string has no \n
Example:

```rust
// this test fails currently
#[test]
fn indented_text_should_have_the_same_number_of_lines_as_the_original_text() {
    let texts = ["foo\nbar", "foo\nbar\n", "foo\nbar\nbaz"];
    for original in texts.iter() {
        let indented = indent(original, "");
        assert_eq!(&indented, original);
    }
}
```

There should be the same number of lines in the input and output (unless the prefix contains a newline).

[Python tests reference](https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Lib/test/test_textwrap.py), [python source reference](https://github.com/python/cpython/blob/a33f2c2bae759fc9d06e1c032fd2026135f2df45/Lib/textwrap.py#L465), can't be implemented the same way in rust without the [`split_inclusive`](https://github.com/rust-lang/rust/issues/72360) feature. (cf. python's [`splitlines`](https://docs.python.org/3/library/stdtypes.html#str.splitlines))

<details>
<summary>Python script passes the same test</summary>

```python
from textwrap import indent

for original in ["foo\nbar", "foo\nbar\n", "foo\nbar\nbaz"]:
    indented = indent(original, "")
    assert original == indented
```
</details>

jRimbault:
Wait for [split_inclusive](https://github.com/rust-lang/rust/issues/72360) to be stable before merging.

closes #207

### DIFF
--- a/src/indentation.rs
+++ b/src/indentation.rs
@@ -18,8 +18,7 @@
 /// ");
 /// ```
 ///
-/// Empty lines (lines consisting only of whitespace) are not indented
-/// and the whitespace is replaced by a single newline (`\n`):
+/// Lines consisting only of whitespace are kept unchanged:
 ///
 /// ```
 /// use textwrap::indent;
@@ -34,7 +33,7 @@
 /// ->Foo
 ///
 /// ->Bar
-///
+///   \t
 /// ->Baz
 /// ");
 /// ```
@@ -45,17 +44,18 @@
 /// ```
 /// use textwrap::indent;
 ///
-/// assert_eq!(indent(" \t  Foo   ", "->"), "-> \t  Foo   \n");
+/// assert_eq!(indent(" \t  Foo   ", "->"), "-> \t  Foo   ");
 /// ```
 pub fn indent(s: &str, prefix: &str) -> String {
     let mut result = String::new();
-    for line in s.lines() {
-        if line.chars().any(|c| !c.is_whitespace()) {
+
+    for line in s.split_inclusive('\n') {
+        if !line.trim().is_empty() {
             result.push_str(prefix);
-            result.push_str(line);
         }
-        result.push('\n');
+        result.push_str(line);
     }
+
     result
 }
 
@@ -138,12 +138,6 @@ pub fn dedent(s: &str) -> String {
 mod tests {
     use super::*;
 
-    /// Add newlines. Ensures that the final line in the vector also
-    /// has a newline.
-    fn add_nl(lines: &[&str]) -> String {
-        lines.join("\n") + "\n"
-    }
-
     #[test]
     fn indent_empty() {
         assert_eq!(indent("\n", "  "), "\n");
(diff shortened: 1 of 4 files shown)


src/indentation.rs
```diff
@@ -34,7 +34,7 @@
 /// ->Foo
 ///
 /// ->Bar
-///
+///   \t
```
mgeisler, line 37: The comment above is now incorrect:

> ... the whitespace is replaced by a single newline (`\n`)

src/indentation.rs
```diff
@@ -45,18 +45,21 @@
...
-            result.push_str(prefix);
-            result.push_str(line);
+    let mut output = String::new();
+
+    for line in s.split_inclusive('\n') {
+        if line.trim_start().len() == 0 {
+            output.push_str(line);
+        } else {
+            output.push_str(prefix);
+            output.push_str(line);
```
mgeisler, line 58: Perhaps it would be a tiny bit simpler to do

```
if (!line.trim().is_empty()) {
  output.push_str(prefix);
}
output.push_str(line);
```

I think that should do the same: conditionally add the `prefix`, unconditionally add the `line` including any trailing whitespace in the `line`.
```

File: src/indentation.rs [hunk 3]
```diff
@@ -141,7 +144,7 @@ mod tests {
     /// Add newlines. Ensures that the final line in the vector also
     /// has a newline.
```
mgeisler, line 145: The comment is outdated now.

src/indentation.rs
```diff
@@ -152,27 +155,29 @@ mod tests {
...
     #[rustfmt::skip]
     fn indent_nonempty() {
-        let x = vec!["  foo",
-                     "bar",
-                     "  baz"];
-        let y = vec!["//  foo",
-                     "//bar",
-                     "//  baz"];
-        assert_eq!(indent(&add_nl(&x), "//"), add_nl(&y));
+        let text = r##"  foo
```
mgeisler, line 158: I tried to use the `add_nl` helper function together with a vector to align the lines in the source code and to make it crystal clear if a string has trailing whitespace. I would prefer to keep this somehow, but I'm open for better ways of doing it :-)

src/indentation.rs
jRimbault: Will fix

src/indentation.rs
jRimbault: Will fix, brainfart here

src/indentation.rs
jRimbault: Will fix

src/indentation.rs
jRimbault: What do you think would be good way to signal trailing whitespace is intentional ?

src/indentation.rs
mgeisler: Well, it's almost the same in both cases :smile:

src/indentation.rs
mgeisler: Hmm... yeah, good question. I guess the most direct way would be to include things like `\n` in the strings, and then join them without adding a separator. So perhaps we can throw out the `add_nl` function and let the tests be explicit:

```rust
        let lines = &[
            "foo\n", //
            "bar\n",
        ];
        let expected = &[
            "  foo\n", //
            "  bar\n",
        ];
        assert_eq!(&indent(&lines.join(""), "  "), &expected.join("")); 
``

the `//` makes `rustfmt` align things consistently so that it is easy to compare the two lists visually. I like that each line is shown in full, with escape characters and everything. Trailing whitespace would then be

```rust
        let lines = &[
            "foo  \n", //
            "bar    ",
        ];
        let expected = &[
            "  foo  \n", //
            "  bar    ",
        ];
        assert_eq!(&indent(&lines.join(""), "  "), &expected.join("")); 
``

If you want, you could try updating the existing tests to use this scheme. It would be a good preparation for this PR.

jRimbault: I'll get on to it on Monday :)

And thanks for the review !

mgeisler: Thanks, the explicit lists look good!

Can I get you to squash all your commits into a single commit with a commit message such as

> Make indent preserve the lack of a trailing \n
>
> Before, indent("foo", " ") would give " foo\n". It now preserves any trailing \n present in the input string. This makes indent behave consistently with dedent. New tests were added to ensure this on a number of corner cases.

You could also mention that we now require Rust version N.NN (when we know this at some point).

jRimbault: Sure :)

mgeisler:
I just noticed that a stabilization PR has been put up for the `split_inclusive` feature: https://github.com/rust-lang/rust/pull/77858 — I'm not sure how long it will take to be reviewed, though.

mgeisler:
This was solved in a simpler way in #279 so I'm closing this now.

PR #217, opened 2020-10-26
Blanket impl for types that deref to WordSplitter

Benjamin-L:
I ran into a situation where it would be useful to use `&hyphenation::Standard` as a splitter. This PR adds a blanket `WordSplitter` impl for anything that derefs to a `WordSplitter`, which should cover cases like this.

### DIFF
--- a/src/splitting.rs
+++ b/src/splitting.rs
@@ -56,6 +56,16 @@ pub trait WordSplitter: std::fmt::Debug {
 #[derive(Clone, Debug)]
 pub struct NoHyphenation;
 
+impl<T> WordSplitter for T
+where
+    T: std::ops::Deref + std::fmt::Debug,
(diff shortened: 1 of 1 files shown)


Benjamin-L:
Wow I'm dumb... I accidentally tested without the `hyphenation` feature. I think this is actually impossible without specialization, so I'm gonna close it.

mgeisler:
Hey @Benjamin-L,

Okay, no worries! Perhaps we can add some more pointers to the `hyphenation` feature so it's easier to find?

Benjamin-L:
That wasn't my problem. I knew about the `hyphenation` feature, and had been using it. I just forgot to include it when I built this change to test.

mgeisler:
I see, glad you figured it out in any case :-)

PR #219, opened 2020-11-05
Allow using trait objects with `fill` & `wrap`.

Cryptjar:
Considering the issue #178 with the current state of `master`, using trait objects with `fill` & `wrap` seems to be a straight forward solution. However, until now, it is not supported by these functions. One problem is that they take the `WrapOptions` by-value, which is incompatible with trait objects.

Digging a bit deeper, it seems that the `WrapIter` already uses the `WrapOptions` by-value, but that trait only defines function that require access via reference. Additionally, considering that the `WrapIter` is typically a rather short-lived entity, it does seem justified to give it owned access to something it only borrows. This issue is also reflected by the implementation for `Options`, which as of now, implements `WrapOptions` only on a reference (i.e. `&Options`).

This PR rectifies this inconsistency by giving `WrapIter` only a reference to some `WrapOptions`. Now, it is no longer required that the given type is sized, and it directly allows for trait objects to be used.

However, the current implementation of this PR, has the drawback that the `usize` as `WrapOptions` must also be given by-reference so a simple `fill("text", 42)` becomes `fill("text", &42)`, it shouldn't cause any issues, but the additional ampersand (`&`) isn't the prettiest.

As shown by a newly added test the following idiom using trait objects becomes now possible due to this PR:

```rust
// Storing the WrapOptions as a boxed trait object. (notice, also `Rc`, `Arc`, and others are possible)
let mut wrapper: Box<dyn WrapOptions> = Box::new(Options::new(80));

assert_iter_eq!(wrap("foo bar baz", &*wrapper), vec!["foo bar baz"]);

// Replace the `Options` with a `usize`
wrapper = Box::new(5);

assert_iter_eq!(wrap("foo bar baz", &*wrapper), vec!["foo", "bar", "baz"]);
```

_This is a follow-up to my early comment, it implements what I called 'proper outer boxing' https://github.com/mgeisler/textwrap/pull/215#issuecomment-722058439_

### DIFF
--- a/examples/multi-layouts.rs
+++ b/examples/multi-layouts.rs
@@ -1,0 +1,99 @@
+use textwrap::{NoHyphenation, Options};
+
+#[cfg(feature = "hyphenation")]
+use hyphenation::{Language, Load, Standard};
+
+// Pretend this is an external crate
+mod library {
+    use textwrap::{Options, WordSplitter};
+
+    #[derive(Debug, Default)]
+    pub struct Layout<'a> {
+        // Use trait-objects which can be easily converted from any concrete `Options`
+        styles: Vec<Box<Options<'a, dyn WordSplitter + 'a>>>,
+    }
+
+    impl<'a> Layout<'a> {
+        pub fn new() -> Self {
+            Default::default()
+        }
+
+        // Similar signature like `wrap` has, so it takes (nearly) everything that `warp` takes.
+        pub fn add<S: WordSplitter + 'a, T: Into<Options<'a, S>>>(&mut self, option: T) {
+            self.styles.push(Box::new(option.into()));
+        }
+
+        pub fn print(&self, text: &str) {
+            // Now we can easily go over all the arbitrary Options and use them for layouting.
+            for opt in self.styles.iter() {
+                println!();
+
+                // the debug output of the hyphenation is very long
+                //println!("Options: {:#?}", opt);
+
+                println!("{:0width$}", 0, width = opt.width);
+
+                // Just use the textwrap functions as usual.
+                // However, we have to first coerce it into a trait-object
+                let dyn_opt: &Options<'a, dyn WordSplitter> = opt;
+                println!("{}", textwrap::fill(text, dyn_opt));
+            }
+        }
+    }
+}
+
+pub fn main() {
+    // pretend we are a user of the library module above
+
+    // Just some options (see below for usage)
+    let some_opt = Options::new(25).initial_indent("----");
+
+    // The struct from the 'library' that we are using
+    let mut layout = library::Layout::new();
+
+    // Add some arbitrary options. We can use here the same as for `fill` & `wrap`.
+
+    // Plain Options
+    layout.add(Options::new(20));
+
+    // usize
+    layout.add(30);
+
+    // Customized Options
+    let opt = Options::new(30);
+    let opt = opt.subsequent_indent("****");
+    layout.add(opt.clone()); // notice, here we pass opt by-value instead of by-reference
+
+    // We can use boxed splitters too (however, we have to coerce the Options)
+    let opt: Options = opt.splitter(Box::new(NoHyphenation));
+    layout.add(opt);
+
+    // We can also pass-in references, however, those need to outlive the local
+    // `layout`, so here, it must be declared before `layout` (drop order).
+    layout.add(&some_opt);
+
+    // If you like, you can download more dictionaries from
+    // https://github.com/tapeinosyne/hyphenation/tree/master/dictionaries
+    // Place the dictionaries in the examples/ directory. Here we
+    // just load the embedded en-us dictionary.
+    #[cfg(feature = "hyphenation")]
+    for lang in &[Language::EnglishUS] {
+        let dictionary = Standard::from_embedded(*lang).or_else(|_| {
+            let path = std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+                .join("examples")
+                .join(format!("{}.standard.bincode", lang.code()));
+            Standard::from_path(*lang, &path)
+        });
+
+        if let Ok(dict) = dictionary {
+            layout.add(Options::with_splitter(25, dict));
+        }
+    }
+
+    let example = "Memory safety without garbage collection. \
+                   Concurrency without data races. \
+                   Zero-cost abstractions.";
+
+    // Printout above text in all different layouts
+    layout.print(example);
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -108,7 +108,7 @@ pub mod core;
 
 /// Holds settings for wrapping and filling text.
 #[derive(Debug, Clone)]
-pub struct Options<'a, S = Box<dyn WordSplitter>> {
+pub struct Options<'a, S: ?Sized = Box<dyn WordSplitter>> {
     /// The width in columns at which the text will be wrapped.
     pub width: usize,
     /// Indentation used for the first line of output.
@@ -125,7 +125,7 @@ pub struct Options<'a, S = Box<dyn WordSplitter>> {
     pub splitter: S,
 }
 
-impl<'a, S> From<&'a Options<'a, S>> for Options<'a, &'a S> {
+impl<'a, S: ?Sized> From<&'a Options<'a, S>> for Options<'a, &'a S> {
     fn from(options: &'a Options<'a, S>) -> Self {
         Self {
             width: options.width,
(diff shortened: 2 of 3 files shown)


mgeisler:
Hey @Cryptjar,

As you note, `fill("my text", &42)` isn't the prettiest API — in fact, I don't like it very much :-) I should post an update on #178: I think the issue there has been solved in `master` already. Let's discuss on the issue!

About changes to `WrapIter`: I've been working on code which rewrites the wrapping code completely. This is to address #126, and I'll put up a PR for that soonish. The `WrapIter` goes away completely and is replaced by a few simpler functions.

Cryptjar:
Now, since you mentioned in #215 that you pushed your rewrite & removal of the `WrapIter`, I revisited this issue and still found it unresolved. I could have just update this PR to the current `master` version, but since you said that you don't like the referencing of `usize`, I actually rewrote this entire PR (which wasn't that much to begin with) to a different approach:

Now, instead of altering the `fill` & `wrap` function signatures, I kept them as-is, and instead add a new generic impl of `WrapOptions` for all references `&T` where `T` already implements `WrapOptions`. This allows now to pass all kind of references to `fill` & `wrap` including reference to trait objects.

Due to the generic impl, I also changed the primary impl of `WrapOptions` for `Options` from its reference to its value (i.e. from `impl WrapOptions for &Options` to `impl WrapOptions for Options`). The generic impl on all references then still allows to pass a reference of `Options` to `fill` & `wrap` (i.e. most (if not all) use-cases should continue to work). But since now `Options` by-value also implements `WrapOptions` some cases where a double reference to an `Options` were required, could be replaced now to a simple reference.

The example from my initial PR comment still works unchanged.

As a side note: this new approach might sound a bit similar to #217, but instead this PR does not introduce a general (blanked) impl for every `T` (every type ever), but only on the specific type 'reference' (`&T`), which in turn just happens to be generic. So it doesn't conflict with other impls.

mgeisler:
Hey @Cryptjar,

> Now, instead of altering the `fill` & `wrap` function signatures, I kept them as-is, and instead add a new generic impl of `WrapOptions` for all references `&T` where `T` already implements `WrapOptions`. This allows now to pass all kind of references to `fill` & `wrap` including reference to trait objects.

Thanks for the rewrite, this sounds great!

> Due to the generic impl, I also changed the primary impl of `WrapOptions` for `Options` from its reference to its value (i.e. from `impl WrapOptions for &Options` to `impl WrapOptions for Options`). The generic impl on all references then still allows to pass a reference of `Options` to `fill` & `wrap` (i.e. most (if not all) use-cases should continue to work). But since now `Options` by-value also implements `WrapOptions` some cases where a double reference to an `Options` were required, could be replaced now to a simple reference.

This also sounds very nice :+1:

However, the changes you propose here overlap with changes proposed by @Koxiaet in #227. I believe you're both basically doing the same thing with blanket implementations for `WrapOptions` for `&T` where `T: WrapOptions`. Are such blanket implementations a normal thing to have?

It seems strange to me — as I wrote on #227, this is similar to implementing the `*` operator :-) Does Rust not already have the auto-deref functionality for this? Or am I mixing things up somehow?

----

Perhaps we can go a step back:

> Now, since you mentioned in #215 that you pushed your rewrite & removal of the WrapIter, I revisited this issue and still found it unresolved.

I'm not 100% sure which "issue" you're referring to here... Is it #178? It's still open since I'm hoping to hear back from the author — as far as I know, it's now possible to embed an `Options` in a struct and so the issue can be closed.

Cryptjar:
> I'm not 100% sure which "issue" you're referring to here... Is it #178? It's still open since I'm hoping to hear back from the author — as far as I know, it's now possible to embed an `Options` in a struct and so the issue can be closed.

Well, what I meant is related to #178 since it showed the symptoms of that _issue_ that I'm referring to. By which I meant, since there is now a trait `WrapOptions` that can be arbitrarily implemented by any user, the most general _thingy_ to describe wrapping parameters is anything that implements `WrapOptions` (so is the argument of `fill` & `wrap` any `T: WrapOptions` which is more than just `usize` & `Options`).

However, if one is writing a library crate (such as it appears to me, to be the case in #178), one might want to store arbitrary user-provided wrapping parameters (aka some `WrapOptions`) in a struct, which might has to be without generics. At this point, in order to allow to store all possible implementations, one way is to use trait-objects (i.e. some `dyn WrapOptions`), which is more general than a simple `Options<Box<dyn WordSplitter>>` (e.g. it would also allow `usize`).

The problem (or issue) is precisely that as of now, once you got a trait-object (e.g. `Box<dyn WrapOptions>`) there is no way to use it with `fill` & `wrap`. And that is what _this PR_ tries to solve.


> However, the changes you propose here overlap with changes proposed by @Koxiaet in #227.

Interestingly, since `Options` lets you freely define all the return values of `WrapOptions`, he got a point, that the trait seems a bit superfluous from a start.


> I believe you're both basically doing the same thing with blanket implementations for `WrapOptions` for `&T` where `T: WrapOptions`. Are such blanket implementations a normal thing to have?
>
> It seems strange to me — as I wrote on #227, this is similar to implementing the `*` operator :-) Does Rust not already have the auto-deref functionality for this? Or am I mixing things up somehow?

_(those are strictly speaking not blanket implementations but generic impls on the generic reference type)_

Rust does have the auto-deref functionality (especially in the context of the `*`-operator), but it only applies to references. For example, if you want to pass a `Box<Options>` into `fill`, you got the problem, that `Box` does not implement `WrapOptions`, no `Deref` involved. But what you really want to do is actually to only pass a reference (`&Box<Options>`) which now thanks to `Deref` can auto-deref to a simple `&Options` and that happens to implement `WrapOptions`.

That is nice and cozy, but it does not help with a trait-object: if you have `Box<Options>` you can coerce it into a `Box<dyn WrapOptions>` and when you have a reference to it, it can auto-deref to a simple `&dyn WrapOptions` but no further. And sadly, as of now there is no implementation of `WrapOptions` for `&dyn WrapOptions` (whereas  `dyn WrapOptions` implements `WrapOptions`, notice the missing reference). And that is what I added in this PR: an implementation of `WrapOptions` for references.

On the other hand, it does not seem that @Koxiaet cares much about trait objects in his PR, but instead his generic implementation seems to have to do with his `From` implementation.

mgeisler:
> The problem (or issue) is precisely that as of now, once you got a trait-object (e.g. `Box<dyn WrapOptions>`) there is no way to use it with `fill` & `wrap`. And that is what _this PR_ tries to solve.

Thanks, I see now!

> > However, the changes you propose here overlap with changes proposed by @Koxiaet in #227.
> 
> Interestingly, since `Options` lets you freely define all the return values of `WrapOptions`, he got a point, that the trait seems a bit superfluous from a start.

Yeah, I agree and I will like to merge the changes there so we can remove `WrapOptions` and work with the `From` trait instead. However, I'm not sure how this plays into the changes in this PR.

Basically, #277 adds a 
```rust
impl From<usize> for Options<'static, HyphenSplitter> {
    fn from(width: usize) -> Self {
        Options::new(width)
    }
}
```

and changes `fill` and `wrap` to take an `into<Options>` argument by value:

```rust
pub fn fill<'a, S: WordSplitter, T: Into<Options<'a, S>>>(text: &str, options: T) -> String {
    // ...
}
```

On the other hand, this PR adds a `WrapOptions` implementation for all `&T` where `T: WrapOptions`:
```rust
impl<T: ?Sized + WrapOptions> WrapOptions for &T {
    // ...
}
```

Can this be combined with the above? This compiles, but I haven't checked if it does the right thing...
```rust
impl<'a, T: ?Sized + Into<Options<'a>>> From<&T> for Options<'a> {
    fn from(other: &T) -> Self {
        other.into()
    }
}
```

mgeisler:
> That is nice and cozy, but it does not help with a trait-object: if you have `Box<Options>` you can coerce it into a `Box<dyn WrapOptions>` and when you have a reference to it, it can auto-deref to a simple `&dyn WrapOptions` but no further. And sadly, as of now there is no implementation of `WrapOptions` for `&dyn WrapOptions` (whereas `dyn WrapOptions` implements `WrapOptions`, notice the missing reference). And that is what I added in this PR: an implementation of `WrapOptions` for references.

This is a really good explanation, thanks! I'm learning a lot here :smile:

mgeisler:
I've merged #227, so the `WrapOptions` trait is no more and things should be a little bit cleaner now. I hope we can then sort out this PR next :-)

Cryptjar:
Sure. That point now is that without the `WrapOptions` trait (that have been removed by #227) there is actually no point in make it into a trait-object. However, without the trait, there can be no user implementations of it either, and the most general configuration parameters can be truly represented by a simple `Options`.

However, for the library use-case I mentioned earlier, it still seems interesting to have some trait-object support such as an `Option<dyn WordSplitter>`. So I rewrote this PR again to now allow this.

This now third approach is what I used to call the 'outer boxing' (i.e. `Box<Option<dyn WordSplitter>>`) in #215, which is similar but different from what I called 'inner boxing' (i.e. `Option<Box<dyn WordSplitter>>`) that already works. To highlight their distinct use-cases, I added now also an example `multi-layouts`, which demonstrates how a library might handle user provided `Options` (i.e. storing them in a `Vec`). This different from what the `interactive` example does (it only has one `Option` where the `splitter` is mutated in-place). So both examples now demonstrate their respective use-case and utilize 'outer boxing' and 'inner boxing' respectively.

mgeisler:
Excellent, thanks for the extensive explanations! They'll be super helpful when I link to the PR in the release notes.

Issue #193, opened 2020-06-29
Non breaking words


truchi:
Hello,

When I run:

```rust
let text = "textwrap: a small library for wrapping text.";

println!(
    "{}",
    textwrap::Wrapper::new(1).break_words(false).fill(text)
);
```

I get the non-expected results:

```
textwrap:
 
a
small
 librar
y
 fo
r
 wrappin
g
 text
.
```

Yet I expect:

```
textwrap:
a
small
library
for
wrapping
text.
```

Same with 2, 3, 4 ...

Thank you!

mgeisler:
Hi @truchi! Oh wow, that looks odd... I'll take a look and try to fix it as soon as possible. Thanks for the report.

truchi:
You are welcome, sir. Thank you for the fix!

Issue #158, opened 2018-12-20
Hyphenation does weird things to subsequent lines


alantrick:
Consider the code:

```rust
use hyphenation;
use hyphenation::Load;
use textwrap;

let us_hyph = hyphenation::Standard::from_embedded(
    hyphenation::Language::EnglishUS).unwrap();
let input = "participation is the key to sucess";
let wrapper = textwrap::Wrapper::new(10);
let output = "participat\nion is the\nkey to\nsucess";
assert_eq!(output, wrapper.fill(input));
let wrapper = textwrap::Wrapper::with_splitter(10, us_hyph);
let output = "participa-\ntion\n \nis the key\nto sucess";
assert_eq!(output, wrapper.fill(input));
// expected output "participa-\ntion is the\nkey to\nsucess";
````

The first wrapper works okay (though it has to split the word in the middle) but the second wrapper with the hyphenator ends up having a blank third line instead of the expected result (in comments).

This might be a bug in the hyphenator library, but I'm not familiar enough with the code to know.

anyakichi:
I am creating the splitter for Japanese and get the same issue.

This should resolve the issue. https://github.com/anyakichi/textwrap/commit/224f0be75ac6e35b5d82a9d637b060502ec669bb

mgeisler:
Hi @anyakichi, thanks so much for debugging this! I'll be super happy to get your patch in a PR so we can fix this bug.

anyakichi:
I've created a PR!

Thank you for useful library.

mgeisler:
Excellent, thanks very much! I'm travelling for a few days, but I'll try to put out a new release as soon as I can.

PR #221, opened 2020-11-08
Reformulate wrapping in terms of boxes, glue, and penalties

mgeisler:
This is a complete rewrite of the core word wrapping functionality. The user-visible change is that `wrap` now returns a `Vec<Cow<'_, str>>` instead of `impl Iterator<Item = Cow<'_, str>>`. In other words, you now get all lines returned to you at once instead of getting an iterator back. Code that simply iterated over the old return value simply need to add `.iter()`, code that already collected the lines into a vector can now do away with this code.  I looked around GitHub for code that uses `textwrap` 
 and most code simply calls `fill` or `wrap`. An example is [the `clap` crate](https://github.com/clap-rs/clap/blob/829d953b489bf4380b02d75706c62b7cbee0f370/src/output/help.rs#L1054).

----

New algorithm: Before, we would step though the input string and (attempt to) keep track of all aspects of the state. This didn't always work (see at least #122, #158, #158, and #193) and it's inflexible.

This commit replaces the old algorithm with a new one which works on a more abstract level. We now first

1. First split the input string into "words". A word is a substring of the original string, including any trailing whitespace.

2. We split each word according to the `WordSplitter`.

3. Optional, if `break_words` is `true`: further spit each word so that it is no longer than the line width.

4. We then simply put the words into lines based on the display width.

This is slower than the previous algorithm. The `fill/1600` benchmark shows that is now takes ~19 microseconds to wrap a 1600 character long string (about 20 lines of terminal text). That is ~8 microseconds longer than before. I think this is still plenty fast, and the new structure makes it easier to reason about the logic.

----

This is a step towards #126: the `wrap_fragments` function could now in principle be used to wrap any kind of opaque "box", and this box could carry formatting information as needed. We can work on abstracting more functionality going forward, probably by making the `Fragment` trait more powerful, e.g., by moving the `break_apart` method from `Word` to `Fragment`.

### DIFF
--- a/examples/interactive.rs
+++ b/examples/interactive.rs
@@ -101,7 +101,7 @@ mod unix_only {
         )?;
         row += 2;
 
-        let mut lines = wrap(text, options).collect::<Vec<_>>();
+        let mut lines = wrap(text, options);
         if let Some(line) = lines.last() {
             // If `text` ends with a newline, the final wrapped line
             // contains this newline. This will in turn leave the
--- a/examples/layout.rs
+++ b/examples/layout.rs
@@ -17,7 +17,7 @@ fn main() {
 
     for width in 15..60 {
         options.width = width;
-        let lines = wrap(example, &options).collect::<Vec<_>>();
+        let lines = wrap(example, &options);
         if lines != prev_lines {
             let title = format!(" Width: {} ", width);
             println!(".{:-^1$}.", title, width + 2);
--- a/src/core.rs
+++ b/src/core.rs
@@ -1,0 +1,588 @@
+//! Building blocks for advanced wrapping functionality.
+//!
+//! The functions and structs in this module can be used to implement
+//! advanced wrapping functionality when the `wrap` and `fill`
+//! function don't do what you want.
+
+use unicode_width::UnicodeWidthChar;
+use unicode_width::UnicodeWidthStr;
+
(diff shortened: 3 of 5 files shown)


mgeisler:
Ah, I should probably explain the PR title a little... the model here is vaguely inspired by the concepts of _boxes_, _glue_, and _penalties_ in TeX. The terms were introduced in the very readable article [Breaking Paragraphs into Lines](http://www.eprg.org/G53DOC/pdfs/knuth-plass-breaking.pdf) from 1981 by Donald E. Knuth and Michael F. Plass. In short, a box is an opaque rectangle on the page, glue is the stretchable whitespace between boxes, and penalties are the extra content inserted at line breaks (such as hyphens). The article describes a line braking algorithm which justifies text while minimizing the stretching of individual lines.

I first wanted to reuse the terminology from the article, but the word `box` is a [reserved keyword](https://doc.rust-lang.org/reference/keywords.html#reserved-keywords) and already has a meaning of "heap allocation". I could have used the word `glue` to refer to the whitespace between words, but since we don't (yet) support justified text, our glue would be rather unflexible. Lastly, the greedy algorithm implemented in `textwrap` does not try to minimize anything except the total number of lines — reusing the terminology from the article would have been misleading people.

PR #226, opened 2020-11-12
Add a `fill_inplace` function

### LINKED ISSUE #166: Request: Ability to have a String wrapped "in place" using `&mut String`
If I already have a String with a large backbuffer, I'd like the ability to apply wrapping to it by calling something like

```rust
textwrap::wrap_in_place(&mut message, textwrap::termwidth());
```

mgeisler:
The `fill_inplace` function fill the text in-place without reallocation the input string. This is convenient if you already have the input available as a `String`.

My benchmarks show that this is about twice as fast as the normal `fill` function.

Fixes #166.


Lokathor:
Excellent!

PR #227, opened 2020-11-13
Replace WrapOptions with Into<Options>

Koxiaet:
It's fairly confusing to have a struct and a nearly-identical trait. By centering everything around the concrete `Options` type, the relationship between `usize` and `Options` can be made more clear.

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -93,7 +93,7 @@ impl<'a> Word<'a> {
 
         Word {
             word: trimmed,
-            width,
+            width: width,
             whitespace: &word[trimmed.len()..],
             penalty: "",
         }
@@ -125,7 +125,7 @@ impl<'a> Word<'a> {
                 if width > 0 && width + ch_width > line_width {
                     let word = Word {
                         word: &self.word[offset..idx],
-                        width,
+                        width: width,
                         whitespace: "",
                         penalty: "",
                     };
@@ -140,7 +140,7 @@ impl<'a> Word<'a> {
             if offset < self.word.len() {
                 let word = Word {
                     word: &self.word[offset..],
-                    width,
+                    width: width,
                     whitespace: self.whitespace,
                     penalty: self.penalty,
                 };
--- a/src/core.rs
+++ b/src/core.rs
@@ -4,6 +4,7 @@
 //! advanced wrapping functionality when the `wrap` and `fill`
 //! function don't do what you want.
 
+use crate::splitting::WordSplitter;
 use unicode_width::UnicodeWidthChar;
 use unicode_width::UnicodeWidthStr;
 
@@ -28,7 +29,7 @@ fn skip_ansi_escape_sequence<I: Iterator<Item = char>>(ch: char, chars: &mut I) 
             }
         }
     }
-    return false;
+    false
 }
 
 /// A (text) fragment denotes the unit which we wrap into lines.
@@ -92,7 +93,7 @@ impl<'a> Word<'a> {
 
         Word {
             word: trimmed,
-            width: width,
+            width,
             whitespace: &word[trimmed.len()..],
             penalty: "",
         }
@@ -124,7 +125,7 @@ impl<'a> Word<'a> {
                 if width > 0 && width + ch_width > line_width {
                     let word = Word {
                         word: &self.word[offset..idx],
-                        width: width,
+                        width,
                         whitespace: "",
                         penalty: "",
                     };
@@ -139,7 +140,7 @@ impl<'a> Word<'a> {
             if offset < self.word.len() {
                 let word = Word {
                     word: &self.word[offset..],
-                    width: width,
+                    width,
                     whitespace: self.whitespace,
                     penalty: self.penalty,
                 };
@@ -242,16 +243,18 @@ pub fn find_words(line: &str) -> impl Iterator<Item = Word> {
 ///     vec![Word::from("foo-bar")]
 /// );
 /// ```
-pub fn split_words<'a, I, T: crate::WrapOptions>(
+pub fn split_words<'a, I, S: WordSplitter, T: Into<crate::Options<'a, S>>>(
     words: I,
-    options: &'a T,
+    options: T,
 ) -> impl Iterator<Item = Word<'a>>
 where
     I: IntoIterator<Item = Word<'a>>,
 {
+    let options = options.into();
+
     words.into_iter().flat_map(move |word| {
         let mut prev = 0;
-        let mut split_points = options.split_points(&word).into_iter();
+        let mut split_points = options.splitter.split_points(&word).into_iter();
         std::iter::from_fn(move || {
             if let Some(idx) = split_points.next() {
                 let need_hyphen = !word[..idx].ends_with('-');
@@ -509,13 +512,13 @@ mod tests {
 
     #[test]
     fn split_words_no_words() {
-        assert_iter_eq!(split_words(vec![], &80), vec![]);
+        assert_iter_eq!(split_words(vec![], 80), vec![]);
     }
 
     #[test]
     fn split_words_empty_word() {
         assert_iter_eq!(
-            split_words(vec![Word::from("   ")], &80),
+            split_words(vec![Word::from("   ")], 80),
             vec![Word::from("   ")]
         );
     }
@@ -523,7 +526,7 @@ mod tests {
     #[test]
     fn split_words_hyphen_splitter() {
         assert_iter_eq!(
-            split_words(vec![Word::from("foo-bar")], &80),
+            split_words(vec![Word::from("foo-bar")], 80),
             vec![Word::from("foo-"), Word::from("bar")]
         );
     }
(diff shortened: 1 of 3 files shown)


mgeisler:
Hey @Koxiaet, this is cool! Thank you — i agree that this makes things much clearer :-)

Can I get you to take out unrelated changes from your commits and also to squash the two commits into a single self-contained commit? I'll review the change now and try to point out what I mean.

src/core.rs
```diff
@@ -124,7 +125,7 @@ impl<'a> Word<'a> {
                 if width > 0 && width + ch_width > line_width {
                     let word = Word {
                         word: &self.word[offset..idx],
-                        width: width,
```
mgeisler, line 127: This change and the other similar changes are unrelated.

src/lib.rs
```diff
@@ -106,27 +106,8 @@ pub use crate::splitting::{HyphenSplitter, NoHyphenation, WordSplitter};
...
-    /// Allow long words to be broken if they cannot fit on a line.
-    /// When set to `false`, some lines may be longer than `width`.
-    fn break_words(&self) -> bool;
-    /// Find indices where `word` can be split.
-    fn split_points(&self, word: &str) -> Vec<usize>;
-}
-
 /// Holds settings for wrapping and filling text.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
```
mgeisler, line 110: Is this related somehow? As far as i recall, the `Standard` type from the `hyphenation` crate is not `Copy`, so I guess the derived copy implementation won't work then — is there then a reason to have it?

src/lib.rs
```diff
@@ -144,79 +125,27 @@ pub struct Options<'a, S = Box<dyn WordSplitter>> {
...
-        self.subsequent_indent
-    }
-    #[inline]
-    fn break_words(&self) -> bool {
-        self.break_words
-    }
-    #[inline]
-    fn split_points(&self, word: &str) -> Vec<usize> {
-        self.splitter.split_points(word)
+impl<'a, S> From<&'a Options<'a, S>> for Options<'a, &'a S> {
```
mgeisler, line 128: This looks a bit weird to me — it's basically saying that you can go from `&Options` to `Options` easily. This is the same as what `*options` would do, right? So would we avoid it somehow?

src/lib.rs
```diff
@@ -520,7 +449,7 @@ impl<'a, S: WordSplitter> Options<'a, S> {
             initial_indent: self.initial_indent,
             subsequent_indent: self.subsequent_indent,
             break_words: self.break_words,
-            splitter: splitter,
```
mgeisler, line 523: Unrelated.

src/lib.rs
```diff
@@ -580,7 +509,7 @@ pub fn termwidth() -> usize {
 /// ```
 ///
 /// [`wrap`]: fn.wrap.html
-pub fn fill<T: WrapOptions>(text: &str, options: T) -> String {
+pub fn fill<'a, S: WordSplitter, T: Into<Options<'a, S>>>(text: &str, options: T) -> String {
```
mgeisler, line 512: Is the `'a` lifetime related to the elided lifetime for `text`? Could `'a` be turned into `'_` perhaps?

src/lib.rs
```diff
@@ -1149,15 +1078,14 @@ mod tests {
         let green_hello = "\u{1b}[0m\u{1b}[32mHello\u{1b}[0m";
         let blue_world = "\u{1b}[0m\u{1b}[34mWorld!\u{1b}[0m";
         assert_eq!(
-            fill(&(String::from(green_hello) + " " + &blue_world), 6),
-            String::from(green_hello) + "\n" + &blue_world
+            fill(&format!("{} {}", green_hello, blue_world), 6),
```
mgeisler, line 1081: While this is a nice change, it seems unrelated. It would be great to have this as a esparate PR.

src/lib.rs
```diff
@@ -1149,15 +1078,14 @@ mod tests {
...
+            format!("{}\n{}", green_hello, blue_world),
         );
     }
 
     #[test]
     fn cloning_works() {
         static OPT: Options<HyphenSplitter> = Options::with_splitter(80, HyphenSplitter);
-        let opt = OPT.clone();
```
mgeisler, line 1160: I don't think the test is doing any cloning now?

src/splitting.rs
```diff
@@ -47,6 +47,11 @@ impl<S: WordSplitter + ?Sized> WordSplitter for Box<S> {
     }
 }
 */
+impl<T: WordSplitter> WordSplitter for &T {
```
mgeisler, line 50: This is another funny looking construct :-) Is this perhaps a use case for the [`Borrow`](https://doc.rust-lang.org/std/borrow/trait.Borrow.html#examples) stdlib trait?

What I mean is that the function that have a type bound on `WordSplitter` could perhaps use a bound which allows them to accept both a `WordSplitter` and a borrowed `WordSplitter`.

mgeisler (changes requested):

src/lib.rs
Koxiaet: The reasoning was that it allows `Config` to be `Copy` in many places, which avoids users having to pass references to it when a copy would be fine.

src/lib.rs
Koxiaet: It's different because it also takes a reference to `S` which removes the need for it to be `Clone`. I also wanted to support passing `&Config` to the functions as it avoids breaking `split_words` too much, which used to take an `&impl WrapOptions`.

src/lib.rs
Koxiaet: Unfortunately not, Rust requires explicit lifetimes here.

src/splitting.rs
Koxiaet: It's a fairly common pattern - see [`AsRef`](https://doc.rust-lang.org/stable/std/convert/trait.AsRef.html#impl-AsRef%3CU%3E), [`Display`](https://doc.rust-lang.org/stable/std/fmt/trait.Display.html#impl-Display-82), [`Hash`](https://doc.rust-lang.org/stable/std/hash/trait.Hash.html#impl-Hash-59), [`Fn`](https://doc.rust-lang.org/stable/std/ops/trait.Fn.html#impl-Fn%3CA%3E). It's also required for the implementation of `From<&Options<'_, S>>` for `Options<'_, &S>` (notice it produces an `&S`, which allows `S` to not be `Clone`).

src/lib.rs
mgeisler: Ah, I hadn't spotted the `S` to `&'a S` change :-)

As for breaking `split_words` — feel free to break it all you want, it's a brand new function and I'm happy to hammer out a better signature before I release it to the wild.

src/lib.rs
mgeisler: Okay, no worries!

src/splitting.rs
mgeisler: Thanks for those references to the stdlib! This makes me feel like we're in good company :-)

src/lib.rs
Koxiaet: `wrap` internally calls `split_words` multiple times. If this was not provided then it would have to (a) clone the `Config` each time and bound `S: Clone` or (b) manually write out the body of this impl, both of which this impl seems preferable to.

That being said, I agree that this impl it is a bit weird. If ergonomics was not an issue all the functions would take `&Options<'_, S>`. However, we want to support `usize`, `Options<'_, S>` _and_ `&Options<'_, S>`. But it's hard to find anything that they all have in common:
* `usize` supports conversion from a reference or value to owned options.
* `Options<'_, S>` supports conversion from a reference to shared options or a value to owned options, or a reference to owned options if we include conversions to `Options<'_, &S>`.
* `&Options<'_, S>` supports conversion from a reference or value to shared options, or to owned options if we include conversions to `Options<'_, &S>`.

The current solution converts a value to owned options, as that's the only thing supported by all three. But you do end up with weird double-references sometimes.

mgeisler: Thanks for that overview, makes this much clearer! It seems I've managed to come up with a bit of a "interesting" API with my weird insistence that `wrap` and `fill` can take both `usize` and some-form-of-options (don't care much here) :smile: 

As for double references... I'm not so concerned with them as long as they're internal to `textwrap` — in particular functions in `textwrap::core` which I consider a place for advanced functionality.

mgeisler:
I'm just trying to understand who the changes here affect the changes in #219 — I hope we can merge both soon and get a nice cleaned up API overall.

src/lib.rs
mgeisler: We'll still need to treat it as non-`Copy` internally, since we cannot rely on `Copy` being derived. This makes me think that it should at least go in a separate PR. I'm still a bit wary of offering one API for users who use a `Copy` type as the `WordSplitter` and a different API for other users — it might be surprising to people is their code stops compiling with the `hyphenation` feature. On the other hand, they might never use that feature and then it's indeed nicer for people if they can just copy around their `Options` as they like.

So yeah, please spin this off into a separate PR which we can then merge in parallel with this.

mgeisler (approved):

mgeisler:
Thanks for the updates and explanations!

Issue #247, opened 2020-12-06
Integer size and overflow when calculating penalty


robinkrahl:
When testing the optimal-fit algorithm, I encountered a panic during the cost calculations:

```
thread 'main' panicked at 'attempt to multiply with overflow', .../textwrap-0.13.0/src/core.rs:710:21
```

I fixed this by multiplying my mm widths with the factor 100 instead of 1000, which should still be a sufficient precision.  But looking at the cost calculations, I have two follow-up questions:

- Why are the costs stored as a `i32` instead of a `usize`?  They are generated by adding positive `usize` values, positive integer constants and other costs, so `usize` should work too.
- How should overflows be handled when calculating the penalty?  I changed the `i32` type to `usize`, causing an integer overflow at the same location for the `max_width` test case that sets the line width to `usize::MAX`.  This is currently working, probably because the huge value `target_width - line_width` is cropped when casting to `i32` before updating the cost.  Shouldn’t the cost calculation use saturating additions and multiplications if large line widths are supported?

mgeisler:
Hey @robinkrahl,

Uh oh... thanks for checking the boundaries! I did some back of the envelope calculations and figured that `gap * gap` should fit in an `i32`, at least for normal line lengths in a terminal.

> * Why are the costs stored as a i32 instead of a usize? They are generated by adding positive usize values, positive integer constants and other costs, so usize should work too.

Yes, I think you're right... while developing it, I was playing negative penalties too, just to see what the effect was. I'm afraid there's no further cleverness behind the numbers than that :smile: 

> Shouldn’t the cost calculation use saturating additions and multiplications if large line widths are supported?

I am unfortunately not 100% sure if it works to return a large fixed value such as `i32::MAX`. The [Python implementation of the SMAWK algorithm](https://github.com/jfinkels/PADS/blob/master/pads/smawk.py#L96-L101) says

> `Matrix(i,j)` should always return a value, rather than raising an
> exception, even for `j` larger than the range we expect to compute.
> If `j` is out of range, a suitable value to return that will not
> violate concavity is `Matrix(i,j) = -i`.  It will not work correctly
> to return a flag value such as `None` for large `j`, because the ties
> formed by the equalities among such flags may violate concavity.

[...]

Issue #224, opened 2020-11-10
Wrap multiple strings into a single paragraph (i.e. keep state while wrapping)


swsnr:
Hello,

I'd like to wrap paragraph text from a Markdown document parsed with pulldown_cmark. 

This library feeds me a stream of events which intersperses the text of a paragraph with other events about formatting, eg:

1. `Text(Lorem ipsum dolor sit )`
2. `Softwrap`
3. `Text(amet, )`
4. `Start(Bold)` 
5. `Text(consetetur)`
6. `End(Bold)` 
7. `Text(sadipscing elitr)`

To wrap all `Text` into a single paragraph I'd currently need to wrap each `Text` individually into a sequence of lines, measure the length of the last line, and then use that length (plus an extra space) as `initial_indent` for the next `Text` event, and so on…

This is certainly possible but not exactly convenient and I can't profit from textwrap's work, which already measured the length of the text.

Could textwrap provide a `ParagraphWrapper` which keeps state of wrapping?

mgeisler:
Hi @lunaryorn,

I'm not 100% sure what you're trying to do, but are you perhaps translating the events into console text? With bold and underlined text via ANSI escape codes? Then this might be useful: https://github.com/mgeisler/textwrap/issues/140#issuecomment-418199515.

lunaryorn:
@mgeisler No, In fact I'm trying to wrap just the text itself because I'm already handling events differently.

I've made an example in playpen: https://play.integer32.com/?version=stable&mode=debug&edition=2018&gist=dd4ab26f4bbed27a3b432a764bded213

Essentially I'd like to have a mutable wrapper which keeps wrapping state across invocations of `wrap_iter`.

mgeisler:
Ah, thank you for the example — that was super helpful!

I've rewritten the wrapping algorithm recently (#221) and I am wondering how the new system can be adapted to your use-case... right now the new top-level `wrap` function returns a `Vec` of strings. However, the underlying wrapping could be changed into something which is iterative instead of being single-shot.

If we imagine we have a `State` object with the current state, then one could imagine doing
```rust
while let Some(text) = some_source_of_text_fragments {
    state.add_text(&text);
    //

    if let Some(line) = state.next_complete_line() {
      // do something with the line
    }
}

if let Some(last) = state.last_line() {
  // do something with the incomplete final line
}
```

So this object would be very much like an iterator, but it would only yield complete lines — and there would then be an extra method to call to get the final incomplete line.

[...]

PR #257, opened 2020-12-19
Implement wrapping functions as iterators

Kestrer:
This is the first mergeable PR using the ideas from #244.

Motivation:
- `no_std`. The current implementation of `wrap_first_fit` does not require allocation at any stage, and so can be used in `no_std` contexts if we support that in the future.
- Avoiding allocations for the return value, for efficency/performance reasons.
- Wrapping without creating a slice. `wrap_first_fit` and `wrap_optimal_fit` now do not require the fragments to be stored contiguously, and instead simply require any iterator over fragments. This is done without losing any performance (in theory at least). This allows piping the result of splitting functions directly into wrapping functions without collecting into a vector first.
- Owned data. It is more flexible to return owned fragments from these functions rather than subslices, as with slices it forces the data to be immutable.

Implementation notes:
- I had to implement `Fragment` for `&F` so that passing a slice into `wrap_first_fit` won't break (as iterating over a slice produces references).
- The `WrapFirstFit` and `WrapOptimalFit` iterators iterate over `(F, bool)` tuples where the bool indicates whether it's the last fragment of its line; from experience of porting this library's code and tests to the new system it is about equally easy to use.

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -32,6 +32,8 @@
 
 use crate::{Options, WordSplitter};
 use std::cell::RefCell;
+use std::fmt::{self, Debug, Formatter};
+use std::iter::FusedIterator;
 use unicode_width::UnicodeWidthChar;
 
 /// The CSI or “Control Sequence Introducer” introduces an ANSI escape
@@ -438,7 +440,7 @@ pub enum WrapAlgorithm {
 /// ```
 /// use textwrap::core::{wrap_first_fit, Fragment};
 ///
-/// #[derive(Debug)]
+/// #[derive(Debug, Clone, Copy)]
 /// struct Task<'a> {
 ///     name: &'a str,
 ///     hours: usize,   // Time needed to complete task.
@@ -465,20 +467,32 @@ pub enum WrapAlgorithm {
 ///     Task { name: "Bathrooms",   hours: 2, sweep: 1, cleanup: 2 },
 /// ];
 ///
-/// fn assign_days<'a>(tasks: &[Task<'a>], day_length: usize) -> Vec<(usize, Vec<&'a str>)> {
+/// fn assign_days<'a>(
+///     tasks: impl IntoIterator<Item = Task<'a>>,
+///     day_length: usize
+/// ) -> Vec<(usize, Vec<&'a str>)> {
 ///     let mut days = Vec::new();
-///     for day in wrap_first_fit(&tasks, |i| day_length) {
-///         let last = day.last().unwrap();
-///         let work_hours: usize = day.iter().map(|t| t.hours + t.sweep).sum();
-///         let names = day.iter().map(|t| t.name).collect::<Vec<_>>();
-///         days.push((work_hours - last.sweep + last.cleanup, names));
+///
+///     let mut task_names = Vec::new();
+///     let mut hours = 0;
+///
+///     for (task, last) in wrap_first_fit(tasks, std::iter::repeat(day_length)).terminate_eol() {
+///         task_names.push(task.name);
+///         hours += task.hours;
+///
+///         if last {
+///             days.push((hours + task.cleanup, std::mem::take(&mut task_names)));
+///             hours = 0;
+///         } else {
+///             hours += task.sweep;
+///         }
 ///     }
 ///     days
 /// }
 ///
 /// // With a single crew working 8 hours a day:
 /// assert_eq!(
-///     assign_days(&tasks, 8),
+///     assign_days(tasks.iter().copied(), 8),
 ///     [
 ///         (7, vec!["Foundation"]),
 ///         (8, vec!["Framing", "Plumbing"]),
@@ -491,7 +505,7 @@ pub enum WrapAlgorithm {
 ///
 /// // With two crews working in shifts, 16 hours a day:
 /// assert_eq!(
-///     assign_days(&tasks, 16),
+///     assign_days(tasks.iter().copied(), 16),
 ///     [
 ///         (14, vec!["Foundation", "Framing", "Plumbing"]),
 ///         (15, vec!["Electrical", "Insulation", "Drywall", "Floors"]),
@@ -502,25 +516,228 @@ pub enum WrapAlgorithm {
 ///
 /// Apologies to anyone who actually knows how to build a house and
 /// knows how long each step takes :-)
-pub fn wrap_first_fit<T: Fragment, F: Fn(usize) -> usize>(
-    fragments: &[T],
-    line_widths: F,
-) -> Vec<&[T]> {
-    let mut lines = Vec::new();
-    let mut start = 0;
-    let mut width = 0;
-
-    for (idx, fragment) in fragments.iter().enumerate() {
-        let line_width = line_widths(lines.len());
-        if width + fragment.width() + fragment.penalty_width() > line_width && idx > start {
-            lines.push(&fragments[start..idx]);
-            start = idx;
-            width = 0;
+pub fn wrap_first_fit<F, W>(fragments: F, line_widths: W) -> WrapFirstFit<F::IntoIter, W::IntoIter>
+where
+    F: IntoIterator,
+    <F as IntoIterator>::Item: Fragment,
+    W: IntoIterator<Item = usize>,
+{
+    WrapFirstFit {
+        inner: WrapFirstFitInner {
+            fragments: fragments.into_iter(),
+            line_widths: line_widths.into_iter(),
+            glue: 0,
+            remaining_width: None,
         }
-        width += fragment.width() + fragment.whitespace_width();
+        .peekable(),
+        terminating_eol: false,
     }
-    lines.push(&fragments[start..]);
-    lines
+}
+
+/// Iterator for [`wrap_first_fit`].
+#[must_use = "iterators are lazy and do nothing unless consumed"]
+pub struct WrapFirstFit<F, W>
+where
+    F: Iterator,
+    <F as Iterator>::Item: Fragment,
+    W: Iterator<Item = usize>,
+{
+    inner: std::iter::Peekable<WrapFirstFitInner<F, W>>,
+    terminating_eol: bool,
+}
+
+impl<F, W> WrapFirstFit<F, W>
+where
+    F: Iterator,
+    <F as Iterator>::Item: Fragment,
+    W: Iterator<Item = usize>,
+{
+    /// Add a terminating EOL (i.e, have the last fragment return `true` in its tuple).
+    pub fn terminate_eol(self) -> Self {
+        Self {
+            terminating_eol: true,
+            ..self
+        }
+    }
+}
+
+impl<F, W> Iterator for WrapFirstFit<F, W>
+where
+    F: Iterator,
+    <F as Iterator>::Item: Fragment,
+    W: Iterator<Item = usize>,
+{
+    type Item = (<F as Iterator>::Item, bool);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let (_, fragment) = self.inner.next()?;
+        let eol = self
+            .inner
+            .peek()
+            .map(|&(newline, _)| newline)
+            .unwrap_or(self.terminating_eol);
+        Some((fragment, eol))
+    }
+}
+impl<F, W> FusedIterator for WrapFirstFit<F, W>
+where
+    F: FusedIterator,
+    <F as Iterator>::Item: Fragment,
+    W: FusedIterator<Item = usize>,
+{
+}
+
+// Manual Debug impl so that `<F as Iterator>::Item: Debug`.
+impl<F, W> Debug for WrapFirstFit<F, W>
+where
+    F: Iterator + Debug,
+    <F as Iterator>::Item: Fragment + Debug,
+    W: Iterator<Item = usize> + Debug,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("WrapFirstFit")
+            .field("inner", &self.inner)
+            .finish()
+    }
+}
+
+// Manual Clone impl so that `<F as Iterator>::Item: Clone`.
+impl<F, W> Clone for WrapFirstFit<F, W>
+where
+    F: Iterator + Clone,
+    <F as Iterator>::Item: Fragment + Clone,
+    W: Iterator<Item = usize> + Clone,
+{
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+            terminating_eol: self.terminating_eol,
+        }
+    }
+}
+
+/// Inner iterator used in [`WrapFirstFit`].
+///
+/// This iterates over tuples of whether the fragment starts on a new line, and the fragment
+/// itself.
+struct WrapFirstFitInner<F, W>
+where
+    F: Iterator,
+    <F as Iterator>::Item: Fragment,
+    W: Iterator<Item = usize>,
+{
+    fragments: F,
+    line_widths: W,
+    /// The glue of the previous fragment on the line.
+    glue: usize,
+    /// The remaining width in the current line. `None` if the line overflows, so not even a
+    /// zero-width fragment can fit on it (unlike `Some(0)` where it can).
+    remaining_width: Option<usize>,
+}
+
+impl<F, W> Iterator for WrapFirstFitInner<F, W>
+where
+    F: Iterator,
+    <F as Iterator>::Item: Fragment,
+    W: Iterator<Item = usize>,
+{
+    type Item = (bool, <F as Iterator>::Item);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let fragment = self.fragments.next()?;
+
+        let new_remaining_width = self
+            .remaining_width
+            .and_then(|width| width.checked_sub(self.glue))
+            .and_then(|width| width.checked_sub(fragment.width()))
+            // Require the penalty to fit on the line; although in theory this can provide
+            // suboptimal line wrapping, it avoids infinite lookahead and backtracking.
+            .filter(|&width| width >= fragment.penalty_width());
+
+        self.glue = fragment.whitespace_width();
+
+        Some(match new_remaining_width {
+            // The line has not overflowed.
+            Some(remaining_width) => {
+                self.remaining_width = Some(remaining_width);
+                (false, fragment)
+            }
+            // The line has overflowed; move the fragment to the next line.
+            None => {
+                let line_width = match self.line_widths.next() {
+                    Some(width) => width,
+                    None => {
+                        self.remaining_width = None;
+                        return None;
+                    }
+                };
+
+                let new_remaining_width = line_width
+                    .checked_sub(fragment.width())
+                    .filter(|&width| width >= fragment.penalty_width());
+                self.remaining_width = new_remaining_width;
+                (true, fragment)
+            }
+        })
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (min_fragments, _) = self.fragments.size_hint();
+        let (min_extra_lines, max_extra_lines) = self.line_widths.size_hint();
+        (
+            // In the worst-case (i.e. fewest fragments yielded) scenario each fragment will take up
+            // one line.
+            std::cmp::min(min_extra_lines, min_fragments),
+            if max_extra_lines == Some(0) && self.remaining_width.is_none() {
+                // If there are no more lines and there is no more space on the current line, we
+                // know that the iterator is finished.
+                Some(0)
+            } else {
+                // Otherwise, there can be an unlimited number of lines.
+                None
+            },
+        )
+    }
+}
+impl<F, W> FusedIterator for WrapFirstFitInner<F, W>
+where
+    F: FusedIterator,
+    <F as Iterator>::Item: Fragment,
+    W: FusedIterator<Item = usize>,
+{
+}
+
+// Manual Debug impl so that `<F as Iterator>::Item: Debug`.
+impl<F, W> Debug for WrapFirstFitInner<F, W>
+where
+    F: Iterator + Debug,
+    <F as Iterator>::Item: Fragment + Debug,
+    W: Iterator<Item = usize> + Debug,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("WrapFirstFitInner")
+            .field("fragments", &self.fragments)
+            .field("line_widths", &self.line_widths)
+            .field("glue", &self.glue)
+            .field("remaining_width", &self.remaining_width)
+            .finish()
+    }
+}
+// Manual Clone impl so that `<F as Iterator>::Item: Clone`.
+impl<F, W> Clone for WrapFirstFitInner<F, W>
+where
+    F: Iterator + Clone,
+    <F as Iterator>::Item: Fragment + Clone,
+    W: Iterator<Item = usize> + Clone,
+{
+    fn clone(&self) -> Self {
+        Self {
+            fragments: self.fragments.clone(),
+            line_widths: self.line_widths.clone(),
+            glue: self.glue,
+            remaining_width: self.remaining_width,
+        }
+    }
 }
 
 /// Cache for line numbers. This is necessary to avoid a O(n**2)
--- a/src/core.rs
+++ b/src/core.rs
@@ -953,12 +953,14 @@ where
         // The last fragment of the range fragments[i..j].
         let last_fragment = &fragments[j - 1];
 
+        // The cumulative width before fragments[i]. This is the cumulative width before the
+        // fragment `i`, or if `i` is the first fragment, zero.
+        let start_cu_width = i.checked_sub(1).map(|i| fragments[i].cu_width).unwrap_or(0);
+
         // Compute the width of a line spanning fragments[i..j] in
         // constant time. We need to adjust widths[j] by subtracting
         // the whitespace of fragment[j-i] and then add the penalty.
-        let line_width = last_fragment.cu_width
-            - i.checked_sub(1).map(|i| fragments[i].cu_width).unwrap_or(0)
-            - last_fragment.whitespace_width
+        let line_width = last_fragment.cu_width - start_cu_width - last_fragment.whitespace_width
             + last_fragment.penalty_width;
 
         // We compute cost of the line containing fragments[i..j]. We
--- a/src/core.rs
+++ b/src/core.rs
@@ -547,7 +547,6 @@ where
             fragments: fragments.into_iter(),
             line_widths,
             line_number: 0,
-            glue: 0,
             remaining_width: None,
         }
         .peekable(),
@@ -659,8 +658,6 @@ struct WrapFirstFitInner<F, W> {
     fragments: F,
     line_widths: W,
     line_number: usize,
-    /// The glue of the previous fragment on the line.
-    glue: usize,
     /// The remaining width in the current line. `None` if the line overflows, so not even a
     /// zero-width fragment can fit on it (unlike `Some(0)` where it can).
     remaining_width: Option<usize>,
@@ -676,35 +673,31 @@ where
 
     fn next(&mut self) -> Option<Self::Item> {
         let fragment = self.fragments.next()?;
+        let fragment_width = fragment.width();
 
-        let new_remaining_width = self
+        let inline_remaining_width = self
             .remaining_width
-            .and_then(|width| width.checked_sub(self.glue))
-            .and_then(|width| width.checked_sub(fragment.width()))
+            .and_then(|width| width.checked_sub(fragment_width))
             // Require the penalty to fit on the line; although in theory this can provide
             // suboptimal line wrapping, it avoids infinite lookahead and backtracking.
             .filter(|&width| width >= fragment.penalty_width());
 
-        self.glue = fragment.whitespace_width();
-
-        Some(match new_remaining_width {
+        match inline_remaining_width {
             // The line has not overflowed.
             Some(remaining_width) => {
-                self.remaining_width = Some(remaining_width);
-                (false, fragment)
+                self.remaining_width = remaining_width.checked_sub(fragment.whitespace_width());
             }
             // The line has overflowed; move the fragment to the next line.
             None => {
                 let line_width = (self.line_widths)(self.line_number);
                 self.line_number += 1;
 
-                let new_remaining_width = line_width
-                    .checked_sub(fragment.width())
-                    .filter(|&width| width >= fragment.penalty_width());
-                self.remaining_width = new_remaining_width;
-                (true, fragment)
+                self.remaining_width = line_width
+                    .checked_sub(fragment_width)
+                    .and_then(|width| width.checked_sub(fragment.whitespace_width()));
             }
-        })
+        }
+        Some((inline_remaining_width.is_none(), fragment))
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
--- a/src/core.rs
+++ b/src/core.rs
@@ -98,6 +98,18 @@ pub trait Fragment: std::fmt::Debug {
     fn penalty_width(&self) -> usize;
 }
 
+impl<'a, F: Fragment + ?Sized> Fragment for &'a F {
+    fn width(&self) -> usize {
+        (*self).width()
+    }
+    fn whitespace_width(&self) -> usize {
+        (*self).whitespace_width()
+    }
+    fn penalty_width(&self) -> usize {
+        (*self).penalty_width()
+    }
+}
+
 /// A piece of wrappable text, including any trailing whitespace.
 ///
 /// A `Word` is an example of a [`Fragment`], so it has a width,
@@ -393,15 +405,23 @@ pub enum WrapAlgorithm {
 /// use textwrap::core::{find_words, wrap_first_fit, wrap_optimal_fit, Word};
 ///
 /// // Helper to convert wrapped lines to a Vec<String>.
-/// fn lines_to_strings(lines: Vec<&[Word<'_>]>) -> Vec<String> {
-///     lines.iter().map(|line| {
-///         line.iter().map(|word| &**word).collect::<Vec<_>>().join(" ")
-///     }).collect::<Vec<_>>()
+/// fn lines_to_strings<'a>(words: impl Iterator<Item = (&'a Word<'a>, bool)>) -> Vec<String> {
+///     let mut lines = Vec::new();
+///     let mut line = String::new();
+///     for (word, eol) in words {
+///         line.push_str(&word);
+///         if eol {
+///             lines.push(std::mem::take(&mut line));
+///         } else {
+///             line.push(' ');
+///         }
+///     }
+///     lines
 /// }
 ///
 /// let text = "These few words will unfortunately not wrap nicely.";
 /// let words = find_words(text).collect::<Vec<_>>();
-/// assert_eq!(lines_to_strings(wrap_first_fit(&words, |_| 15)),
+/// assert_eq!(lines_to_strings(wrap_first_fit(&words, |_| 15).terminate_eol()),
 ///            vec!["These few words",
 ///                 "will",  // <-- short line
 ///                 "unfortunately",
@@ -409,7 +429,7 @@ pub enum WrapAlgorithm {
 ///                 "nicely."]);
 ///
 /// // We can avoid the short line if we look ahead:
-/// assert_eq!(lines_to_strings(wrap_optimal_fit(&words, |_| 15)),
+/// assert_eq!(lines_to_strings(wrap_optimal_fit(&words, |_| 15).terminate_eol()),
 ///            vec!["These few",
 ///                 "words will",
 ///                 "unfortunately",
@@ -476,7 +496,7 @@ pub enum WrapAlgorithm {
 ///     let mut task_names = Vec::new();
 ///     let mut hours = 0;
 ///
-///     for (task, last) in wrap_first_fit(tasks, std::iter::repeat(day_length)).terminate_eol() {
+///     for (task, last) in wrap_first_fit(tasks, |_| day_length).terminate_eol() {
 ///         task_names.push(task.name);
 ///         hours += task.hours;
 ///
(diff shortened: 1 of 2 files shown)


Koxiaet:
Oh no, performance has regressed for `wrap_first_fit` by up to 50% :(. Optimal fit performance hasn't changed at least (in a couple benchmarks it even improved slightly).

There are three places where we allocate vectors that could be replaced with iterators after this change: in `break_words`, after we call `wrap_first_fit`/`wrap_optimal_fit` in `wrap`, and when generating the list of lines in `wrap`. Hopefully doing that will improve performance - but I'll leave it for another PR.

mgeisler:
Hey @Koxiaet, thanks for putting this up. It'll take a bit to digest it with Christmas coming up :smile: 

i'll try and test it and drop some comments over the next few days.

mgeisler:
> Oh no, performance has regressed for `wrap_first_fit` by up to 50% :(. Optimal fit performance hasn't changed at least (in a couple benchmarks it even improved slightly).

When i was playing with this in the past, I remember that the trade-off between using vectors and iterators wasn't always clear. Sometimes allocating a few vectors and iterating over them is faster than doing everything with iterators. I'm not 100% sure how it works, but I guess it's a matter of having more complex code vs using more memory upfront.

> There are three places where we allocate vectors that could be replaced with iterators after this change: in `break_words`, after we call `wrap_first_fit`/`wrap_optimal_fit` in `wrap`, and when generating the list of lines in `wrap`. Hopefully doing that will improve performance - but I'll leave it for another PR.

It would be interesting to see the impact of this on the performance. But please separate this into other commits so that you can easily move it to a separate PR.

src/core.rs
```diff
@@ -730,20 +995,64 @@ pub fn wrap_optimal_fit<'a, T: Fragment, F: Fn(usize) -> usize>(
...
+            .map(|(i, data)| {
+                let eol = i + 1 == line_start;
+                if eol {
+                    line_start = minima[line_start].0;
+                }
+                (data.fragment, eol)
+            })
+            .collect(),
+    }
+}
```
mgeisler, line 1014: Is the `WrapOptimalFit` struct needed here? It seems to simply contain a list of tuples and I guess another `.map` here could make it unnecessary?

src/core.rs
```diff
@@ -671,30 +906,60 @@ const HYPHEN_PENALTY: i32 = 25;
...
-    let mut widths = Vec::with_capacity(fragments.len() + 1);
-    let mut width = 0;
-    widths.push(width);
-    for fragment in fragments {
-        width += fragment.width() + fragment.whitespace_width();
-        widths.push(width);
+pub fn wrap_optimal_fit<F, W>(
+    fragments: F,
+    line_widths: W,
+) -> WrapOptimalFit<<F as IntoIterator>::Item>
```
mgeisler, line 912: Is there a reason not to let the functions return `impl Iterator` instead of concrete structs? That should slim down the API, just like I did in #201.

src/core.rs
Koxiaet: The key thing is that `WrapOptimalFit::next` calls `.pop()` on the underlying vector, which means that it reverses the order of iteration (which is essential for the algorithm to work). To still support getting the vector directly with `Vec::reverse` like we had before `WrapOptimalFit` also has a `.into_vec()` method.

src/core.rs
Koxiaet: For both of the iterators I have a `.terminate_eol()` function, which I think is very useful - I use both (terminating and non-terminating) variants in different parts of the library. It could be added as an extra parameter to the wrapping functions of course, but it makes it harder to call the functions and it is not obvious what the `true` in something like `wrap_optimal_fit(fragments, widths, true)` would mean - on the other hand `wrap_optimal_fit(fragments, widths).terminate_eol()` is very clear.

Also, `WrapOptimalFit::into_vec` is useful for efficiency, and it couldn't exist with `impl Iterator`.

src/core.rs
```diff
@@ -671,30 +906,60 @@ const HYPHEN_PENALTY: i32 = 25;
...
+        // The last fragment of the range fragments[i..j].
+        let last_fragment = &fragments[j - 1];
+
         // Compute the width of a line spanning fragments[i..j] in
         // constant time. We need to adjust widths[j] by subtracting
         // the whitespace of fragment[j-i] and then add the penalty.
-        let line_width = widths[j] - widths[i] - fragments[j - 1].whitespace_width()
-            + fragments[j - 1].penalty_width();
+        let line_width = last_fragment.cu_width
+            - i.checked_sub(1).map(|i| fragments[i].cu_width).unwrap_or(0)
```
mgeisler, line 960: I think this deserves a comment :-) Or rather, please pull it out into a local variable.

I guess you're basically differentiating between `i == 0` (which is mapped to `0`) and `i > 0` which is mapped to `fragments[i].cu_width`? Why is this needed? Before I computed `widths` so that I could always index into it with `i` and `j`, but I guess that changed?

mgeisler:
> * `no_std`. The current implementation of `wrap_first_fit` does not require allocation at any stage, and so can be used in `no_std` contexts if we support that in the future.

Do you have such a use case? While it would be technically cool to support `no_std`, nobody has asked for such a feature yet. I really like it from the aspect of better performance and less memory consumption, but if it makes things much more complex, then I prefer allocating a few `Vec`s once in a while :-)

src/core.rs
mgeisler: I haven't yet figured out what this `bool` is for :see_no_evil:  but I haven't look super careful at the code yet.

Koxiaet:
No, I do not have a use case, and I agree we shouldn't bother actually supporting it until someone asks. But designing it this way would avoid breaking changes should someone ask. Also I still like iterators over `Vec`s; this is `core` after all - not the primary way to consume the crate - and an extra iterator type isn't that complex.

Koxiaet:
It's worth mentioning that the previous implementation was technically incorrect; if the fragments' widths' had been nearing the integer limit, it would have had all sorts of problems with integer overflow. On the other hand, this version only uses checked arithmetic which does degrade performance but allows it to be more correct. I benchmarked this version using the (incorrect) unchecked arithmetic and performance increased by up to 10%.

mgeisler:
> It's worth mentioning that the previous implementation was technically incorrect; if the fragments' widths' had been nearing the integer limit, it would have had all sorts of problems with integer overflow.

Yeah, I'm fighting with those overflow problems in #247...

My thinking is that `wrap_optimal_fit` should handle fragments with a combined width fitting in an `usize`. This fails today and you're right that an iterator version of `wrap_first_fit` would be able to handle such "over-sized" strings.

Allowing the input string to come from an iterator has been discussed in #224 — not for the purpose of wrapping strings larger than `usize`, but for the purpose of consuming the wrapping input which is yielded bit by bit.

mgeisler:
> No, I do not have a use case, and I agree we shouldn't bother actually supporting it until someone asks. But designing it this way would avoid breaking changes should someone ask.

Yes, good point. I think we agree that a fully iterator-based design is the most flexible and powerful design. My "plan" was roughly to use iterators for all the helper functions and then let `wrap_optimal_fit` switch to a `Vec` while `wrap_first_fit` could continue to use the iterator.

It's of course always a question if the flexible solution is reasonable in terms of complexity and performance. The code already got slower when I introduced the `Fragment`s in #221: the old messy code that did everything in one big loop was about twice as fast as the current code.

> Also I still like iterators over `Vec`s; this is `core` after all - not the primary way to consume the crate - and an extra iterator type isn't that complex.

Yeah, fair enough.

Koxiaet:
I looked at the flamegraph to try and find what was causing the slow down. Here is the flamegraph on the old code:

fg1

And here is the flamegraph on the new code:

fg2

`break_words` still takes up _most_ of the time spent wrapping, but now lots of time is spent collecting `WrapFirstFit` into a vector. Of that time, about half of it isn't labelled, so I can only assume that it's reallocating - which somewhat makes sense, the new vector is much much longer than the old one since it stores an entry for each fragment rather than an entry for each line. And a quarter is running `ptr::write`. Luckily, these won't exist once we remove the collects so I don't have to worry much about it.

But the time spent on `next` is still a bit more than the time that was spent on `wrap_first_fit`. This is mostly due to `Peekable` - maybe removing that would make it go faster? And the time spent in `WrapFirstFitInner::next` itself is tiny, so it's probably not worth trying to optimize that.

Koxiaet:
@mgeisler Please can you finish reviewing this? It's starting to conflict.

mgeisler:
> @mgeisler Please can you finish reviewing this? It's starting to conflict.

Good morning! Yes, I've also had this on my mind the last few days. The conflicts should be trivial: I just moved the `wrap_optimal_fit` function to its own hidden module so that I could make the smawk dependency optional in #261.

You mentioned integer limits above and I've put up #259 to handle this. It's not merged yet since I haven't heard back from @robinkrahl if this is approach is really better than a simple `panic!("Stop using long fragments")` at the start of `wrap_optimal_fit`. But fixing crashes has been my first priority and it's something that I would like to get a release out for soon.

src/core.rs
mgeisler: I looked more at the code and I think I see it now: instead of yielding full lines, your iterators yield _once per `Fragment`_. The `bool` here tells the caller if the `Fragment` you just saw is at the end of a line.

This sounds expensive to me since the caller needs to "touch" every fragment. The existing code allows the caller in `wrap_first_fit` to simply deal with a full line at a time.

mgeisler:
> Luckily, these won't exist once we remove the collects so I don't have to worry much about it.

I'm not sure I understand this? The `benches/linear.rs` benchmarks use `fill` to ensure that the strings from `wrap` are collected into a big `String`. So I think you end up paying for collecting in any case, no? If not in `wrap_first_fit` and `wrap_optimal_fit` like today, then in whoever calls them.

Today the input and output types of the helper functions look like this (where I wrote `Iterator<T>` instead of `Iterator<Item = T>`):

| Function      | Current input    | Output             | Your input           | Output                       |
|---------------|------------------|--------------------|----------------------|------------------------------|
| `find_words`  | `&str`           | `Iterator<Word>`   | `&str`               | `Iterator<Word>`             |
| `split_words` | `Iterator<Word>` | `Iterator<Word>`   | `Iterator<Word>`     | `Iterator<Word>`             |
| `break_words` | `Iterator<Word>` | `Vec<Word>`        | `Iterator<Word>`     | `Vec<Word>`                  |
| `wrap_*_fit`  | `&[Fragment]`    | `Vec<&[Fragment]>` | `Iterator<Fragment>` | `Iterator<(Fragment, bool)>` |

The helpers stay the say, but the `wrap_*_fit` functions now take and return iterators. So an input `&str` goes through these transformations inside of `textwrap:wrap`:

```rust
let line:          &str           = ...;
let words:         Iterator<Word> = find_words(text);
let split_words:   Iterator<Word> = split_words(words);
let broken_words:  Vec<Word>      = break_words(split_words);
let wrapped_lines: Vec<&[Word]>   = wrap_first_fit(&broken_words);
let cow_lines:     Vec<Cow<str>>  = ... // loop over each wrapped line and slice `line`
```

With your changes, the work is shifted a bit since `wrap_*_fit` return an iterator:

```rust
let line:          &str                   = ...;
let words:         Iterator<Word>         = find_words(text);
let split_words:   Iterator<Word>         = split_words(words);
let broken_words:  Vec<Word>              = break_words(split_words);
let wrapped_words: Iterator<(Word, bool)> = wrap_first_fit(&broken_words);
let cow_lines:     Vec<Cow<str>>          = ... // loop over each Word, check if eol is set and slice `line`
```

The `wrap` function is now the one that keeps track of when the lines are done — before this was done by `wrap_*_fit` since they returned slices of words. Put differently, your approach seems to delay the work a little so that more of it is done when the iterator is advanced in `wrap` instead of doing it when words are wrapped in `wrap_*_fit`. Do I understand that correctly?

As a whole, I'm concerned about the 50% slowdown for the otherwise fast `wrap_first_fit` function. My impression is that putting iterators on top of iterators only work well up to a limit. I don't know why the new version is slower — perhaps because you turn the iterator into a vector here:

```rust
        // This currently isn't as efficient as it could be as we collect into a vec even when it
        // isn't necessary.
        let wrapped_words = match options.wrap_algorithm {
            core::WrapAlgorithm::OptimalFit => core::wrap_optimal_fit(broken_words, line_lengths)
                .terminate_eol()
                .into_vec(),
            core::WrapAlgorithm::FirstFit => core::wrap_first_fit(broken_words, line_lengths)
                .terminate_eol()
                .collect(),
        };
```
As you mention, this vector is much bigger than the old per-line vector.

This points to a problem I ran into with iterators: it's super annoying to match up the types between different iterators. Whereas one vector of `Word`s is the same type as any other vector of `Word`s, iterators almost always end up being different types. So we cannot do
```rust
        let wrapped_words = match options.wrap_algorithm {
            core::WrapAlgorithm::OptimalFit => core::wrap_optimal_fit(broken_words, line_lengths),
            core::WrapAlgorithm::FirstFit => core::wrap_first_fit(broken_words, line_lengths),
        };
```
since the two functions will return different iterators — but we can easily assign the results to the same variable when we collect into a vector.

So if `wrap` should return an iterator too, then it would have to be a cleverly designed type which knows about the two wrapping algorithms.

src/core.rs
```diff
@@ -502,30 +536,188 @@ pub enum WrapAlgorithm {
...
+    }
+}
+
+impl<F, W> FusedIterator for WrapFirstFit<F, W>
+where
+    F: FusedIterator,
+    <F as Iterator>::Item: Fragment,
+    W: Fn(usize) -> usize,
+{
+}
```
mgeisler, line 620: This doesn't seem to change anything with regards to performance, so what is it good for?

Koxiaet:
> So I think you end up paying for collecting in any case, no?

You will end up collecting at least once, but at the moment there are three more collects than necessary - `break_words`, `wrap` and `wrap_first_fit` are all collected (or push to a `Vec`) when they could use iterators. That's what I meant by removing the collects.

> The `wrap` function is now the one that keeps track of when the lines are done — before this was done by `wrap_*_fit` since they returned slices of words. Put differently, your approach seems to delay the work a little so that more of it is done when the iterator is advanced in `wrap` instead of doing it when words are wrapped in `wrap_*_fit`. Do I understand that correctly?

I wouldn't consider my `wrap` to do more or less work, it just does it differently. While the code before had to iterate over the lines and then further iterate over each fragment in the line, my code iterates over everything in one loop. There is a bit more state to keep track of as a result, but it's only one number. If you look at the number of lines in `wrap` my implementation is only a couple lines longer than before. And if you look at `fill_inplace`, my version _reduces_ the number of lines of code significantly.

> I don't know why the new version is slower — perhaps because you turn the iterator into a vector here:

This is what I was referring to when I was talking about the collects earlier; I also suspect that.

> It's super annoying to match up the types between different iterators.

> So if wrap should return an iterator too, then it would have to be a cleverly designed type which knows about the two wrapping algorithms.

Yes, agreed. I have a plan to make an enumeration type over `WrapFirstFit` and `WrapOptimalFit`. I also want to add a method on `WrapAlgorithm` that returns this enumeration type given the broken words and line lengths, which might be useful.

src/core.rs
Koxiaet: Since those traits exist in std and can be implemented, why not? They might be useful to a user of the library.

mgeisler: Ah, simply because it's 20 lines of extra code without a clear benefit. This makes it noise in this PR, IMHO. Please sneak it in at a later point — perhaps you have an application outside of textwrap which benefits from it :-)

mgeisler:
> If you look at the number of lines in `wrap` my implementation is only a couple lines longer than before. And if you look at `fill_inplace`, my version _reduces_ the number of lines of code significantly.

Yes, and it's faster than before for small strings:
```
String lengths/fill_inplace/400                                                                             
                        time:   [2.8580 us 2.8652 us 2.8708 us]
                        change: [-12.352% -11.913% -11.510%] (p = 0.00 < 0.05)
String lengths/fill_inplace/2400                                                                             
                        time:   [17.235 us 17.247 us 17.258 us]
                        change: [-0.2181% +0.1055% +0.3754%] (p = 0.51 > 0.05)
String lengths/fill_inplace/6400                                                                             
                        time:   [45.494 us 45.542 us 45.594 us]
                        change: [+0.1729% +0.4338% +0.7480%] (p = 0.00 < 0.05)
```

I generated the above with
```
$ git checkout 0.13.1
$ cargo bench --bench linear /6400 -- -s 0.13.1
$ git checkout Koxiaet-wrap-iterators
$ cargo bench --bench linear /6400 -- -b 0.13.1
```

So that's indeed a nice win :+1:

mgeisler:
> I wouldn't consider my `wrap` to do more or less work, it just does it differently. While the code before had to iterate over the lines and then further iterate over each fragment in the line, my code iterates over everything in one loop. There is a bit more state to keep track of as a result, but it's only one number.

I see that it's roughly doing the same since the tests pass :smile: However, it's still somehow ~30-35% slower than before:

```
String lengths/fill_first_fit/400
                        time:   [7.6960 us 7.7117 us 7.7376 us]
                        change: [+34.473% +34.829% +35.181%] (p = 0.00 < 0.05)
String lengths/fill_first_fit/2400
                        time:   [40.316 us 40.360 us 40.406 us]
                        change: [+32.506% +32.926% +33.325%] (p = 0.00 < 0.05)
String lengths/fill_first_fit/6400
                        time:   [109.63 us 109.71 us 109.81 us]
                        change: [+32.979% +33.342% +33.697%] (p = 0.00 < 0.05)
```

So this means this PR is offering more code :-1:, more public API surface :-1:, more future flexibility :+1:, potential future savings :+1:, and an immediate performance regression :-1:. The balance of this seems to be potential future gains and we pay for this with more complexity now.

I think the way forward is to simplify this change as much as absolutely possible. Rip out anything that isn't essential — send PRs with the uncontroversial parts such as your nice explanation of the cost matrix. You also added a TODO about making `fill_inplace` take a `&mut str`: that seems like a great idea and I don't know why I didn't use that signature already. Push that to separate PRs so that this PR becomes crystal clear and focused.


I cannot expect you to know the history of the library, but as I've mentioned before, the library changed a lot between version 0.12 and 0.13. I hated the way the code had grown to be in version 0.12.1: it was messy, overly complex, and inflexible. On the other hand, I'm so far very happy with the cleanups I've done in the last few months. So this is why I'm pushing back on what seems to be "regressions" compared to these PRs:

* #201: Let functions return "impl Iterator" to hide iterator structs.

    I removed such structs from the API:
    * https://github.com/mgeisler/textwrap/blob/0.12.1/src/lib.rs#L377
    * https://github.com/mgeisler/textwrap/blob/0.12.1/src/lib.rs#L415

    If the explicit return types can be replaced with `-> impl Iterator`, then please do so.

    If you can replace the structs with

    ```rust
    let mut some_iterator_state = "";
    let mut more_state = 0;
    return std::iter::from_fn(move || {
        // lovely self-contained logic
    }
    ```

    then even better! I found the old code to be a maze and I'm afraid it could move in that direction again.

* #221: Reformulate wrapping in terms of words with whitespace and penalties.

    I got rid of a nested iterator struct  here:

    * https://github.com/mgeisler/textwrap/blob/0.12.1/src/lib.rs#L435-L446

    It held a ton of state and though I tried my best to document it, I ended up with page after page of code like this:

    * https://github.com/mgeisler/textwrap/blob/0.12.1/src/lib.rs#L478-L512

    The current code is much more direct while also being much more flexible. I think there are none of these small boiler-plate functions left:

    * https://github.com/mgeisler/textwrap/blob/0.12.1/src/lib.rs#L377-L382
    * https://github.com/mgeisler/textwrap/blob/0.12.1/src/lib.rs#L415-L423
    * https://github.com/mgeisler/textwrap/blob/0.12.1/src/lib.rs#L440-L446
    * https://github.com/mgeisler/textwrap/blob/0.12.1/src/lib.rs#L465-L468

    There was so much repetition in the old code, duplicated docstrings everywhere, etc. I'm not saying you're duplicating things, but every struct ends up being part of the public API and as such it gets a docstring and its own little page on docs.rs.

src/core.rs
```diff
@@ -502,30 +536,188 @@ pub enum WrapAlgorithm {
...
+    <F as Iterator>::Item: Fragment + Clone,
+    W: Fn(usize) -> usize + Clone,
+{
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+            terminating_eol: self.terminating_eol,
+        }
+    }
+}
```
mgeisler, line 650: If this isn't needed, then please drop it. I'm not sure what the semantics would be of cloning an `Iterator` would be? Does advancing one iterator also advance the other?

Koxiaet: Implementing `Clone` for iterators is very common - the standard library does it for nearly every one of its iterator types and itertools does too. I don't see the benefit in _not_ having this - having it could be useful and it's unlikely someone will go out of their way to implement this at any point, so why not make your library slightly more useful?

src/core.rs
Koxiaet: Like I replied in the responses below - this isn't about reducing the number of lines of code to the absolute minimum possible. This isn't code golf. Libraries should be able to grow and shrink to whatever size is necessary for support a complete API for their purpose. "Complete" here means implementing standard library traits where possible, because they could be useful to users when they can't trivially work around it (without performance loss with `.fuse()` or something). Reactive maintainance, while it is the best option for some scenarios, is ineffective here because this impl is so tiny, and it is very unlikely that someone would go out of their way to implement it.

Do I need this impl? No, I do not. I can live fine without it, and likely all your users can. But it does not introduce complexity. Lines of code does not necessarily equal complexity, and these lines of code are very, very simple. The trait bounds are almost identical to above, and there is exactly one line of code that is executed. That's why I think this impl shoud stay - not because there is great benefit, but because there is a little benefit and a tiny cost.

Koxiaet:
So it seems much of the conflict here is over whether to use `impl Iterator` or manual iterator types. I prefer the latter for four reasons.

First of all, an argument very similar to the ones I used in the comments above - lines of code != cost. Complexity is cost. And although manual iterator types introduce many more lines of code, its complexity is identical to `impl Iterator`. Having another type in the API surface is a bit more complex than not having it, but it's very minor, and when there is a function `wrap_first_fit` as well as a struct `WrapFirstFit`, users will know exactly what the struct does and will know of its importance.

Second, being able to name types is useful. Yes, this limitation won't exist in a few years once `impl Trait` in type aliases becomes stable, but it's not stable yet and even when it is it will require extra boilerplate for the user, as well as inconsistent naming of the iterator among different users. This is even relevant for us - in order to make a iterator that enumerates `WrapFirstFit` and `WrapOptimalFit`, we require being able to name those types.

Third, precedent. Once again I will point to the standard library, which never uses `impl Trait` in return position. But it is also a very common pattern to have named iterator types for user crates - the [Rust API guidelines on naming](https://rust-lang.github.io/api-guidelines/naming.html#iterator-type-names-match-the-methods-that-produce-them-c-iter-ty) has a section on this. And just in general named iterators are common, there are many examples in the ecosystem like `unicode-segmentation`, `itertools`, etc.

Fourth, it allows us to implement traits conditionally. A trait impl like "only implement `FusedIterator` when the underlying iterator implements `FusedIterator`" is simply not expressable with `impl Trait`. This is also relevant when implementing `Clone`, which is recommended to implement by [the C-COMMON-TRAITS guideline](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits).

src/core.rs
mgeisler: I see that you're right — the `Clone` and `Debug` implementations are everywhere in the standard library... I hadn't noticed them until now.

The benefit of not providing the implementation is the same as usual: less code to maintain, less constraints on future developments. Less API surface to maintain also results in less noise in the documentation. I consider things like [`option::Iter`](https://doc.rust-lang.org/std/option/struct.Iter.html), [`result::Iter`](https://doc.rust-lang.org/std/result/struct.Iter.html), etc noise. If the stdlib had been written today, I fully expect that it would have hidden all these structs behnd `-> impl Iterator + Debug + ...`.

mgeisler: I guess I just dislike code :smile: The less, the better, in my opinion. I feel better about the code if I know what problem it solves, and I've always tried to avoid solving potential problems that haven't actually materialized themselves yet.

That being said, I guess these implementations aren't so important and I guess they can stay. It would just be nice if you had an example program that needs this code somehow.

mgeisler:
> So it seems much of the conflict here

I think the main issue is that this change enables _potential_ future flexibility and that the cost for this is paid now. The performance regression for `wrap_first_fit` is a regression and I don't see enough gain yet to offset this cost.

I understand that this is a stepping stone towards a more streaming API. In the limit, this could allow wrapping text from an `impl Iterator` and we could return `impl Iterator` as well without materializing a `Vec` at any point (as you've pointed out). This is nice, though it conflicts with a few things:

1. We don't know how expensive this really is.
2. We probably don't really need this since wrapping wtih a `Vec` is super fast.
3. We have a wrapping algorithm which _needs_ a `Vec` anyway...

This makes me say that we need some real use case to pay the price of the performance hit. Ideally, the change should be a performance improvement — which would justify what I see as downsides in terms of code organization and more types in the public API.

I had a streaming API in the past — and crates like clap would simply [wrap a `&str` and return a `String`](https://github.com/clap-rs/clap/blob/a0269a41d4abaf4b0a9ec4f9a059fe62ea0ba3a7/src/output/help.rs#L1047-L1053). Furthermore, https://github.com/clap-rs/clap/issues/1365 mentions that textwrap is already showing up as a largeish dependency of clap, and it would be a shame if textwrap ended up being dropped because of too much complexity.

> Having another type in the API surface is a bit more complex than not having it, but it's very minor

I tried to point to the old code above to show how it was much more annoying to work with — the artificial split between the function returning the iterator, the struct itself, and the `impl Iterator` block moves three things apart which ought to be one. It was a step forward for me when I got rid of that split — and I'm reluctant to reintroduce it.

> Third, precedent. Once again I will point to the standard library, which never uses `impl Trait` in return position. But it is also a very common pattern to have named iterator types for user crates - the [Rust API guidelines on naming](https://rust-lang.github.io/api-guidelines/naming.html#iterator-type-names-match-the-methods-that-produce-them-c-iter-ty) has a section on this. And just in general named iterators are common, there are many examples in the ecosystem like `unicode-segmentation`, `itertools`, etc.

Is this not simply because the code was written long before `-> impl Iterator` was a thing? I would expect all new libraries to not expose the underlying iterator struct.

Koxiaet:
> too much complexity.

I think that's the problem here - we have different definitions of "simple" and "complex". I want this API, because I consider iterators to be simpler than vectors, _even if_ they have more lines of code.

I believe that APIs, in order to be maximally simple, should do one thing and one thing only. An API that returns a sequence of values shouldn't do other things like allocate, it shouldn't prescribe a collection type for the user, it shouldn't require that slices of fragments are what the user needs to pass in - these are all out of scope and over-complicate the API. Iterators on the other hand, while they do need more code to implement, are theoretically simpler because they don't do anything other than their singular purpose. So if there's a slight performance loss or there are more types in the public API, so be it. To me it's far better to have a simple API than a fast one or a small one (unless it's _really_ slow of course :smile:).

This situation is an interesting clash between the two philosophies of "do one thing and do it well" and using as few lines as possible :laughing:. I can understand _why_ you'd subscribe to the latter, I just find it a bit reductionist.

> textwrap is already showing up as a largeish dependency of clap

Hmm, this is somewhat of a problem. Maybe more aggressive feature-flagging is the way to go? I see you're already doing that a bit with the other PRs in this repo.

> moves three things apart which ought to be one

Is having them next to each other in the source not enough? I agree it's not quite as together as if it was all in one function, but it's not that much harder to navigate.

> I would expect all new libraries to not expose the underlying iterator struct.

No, the standard library still prefers named types. For example, the recently added `core::future::ready` and `core::future::pending` APIs return `core::future::Ready` and `core::future::Pending`, even when they could be `async fn`s which would the number of LOC. And those are futures, which even I would usually not support writing named types for.

src/core.rs
mgeisler: Okay, I was hoping there would be some concrete performance benefit of it, apart from "checking the boxes" by implementing the marker traits from the stdlib. I did a few searches on GitHub, but I was unable to find examples of code that exploit the `FusedIterator` trait — `ExactSizeIterator` seems easier to take advantage of and I see people do that here and there.

I agree with you that these few lines are unlikely to cause headaches later.

src/core.rs
mgeisler: > Also, `WrapOptimalFit::into_vec` is useful for efficiency, and it couldn't exist with `impl Iterator`.

Would it be possible to implement `Into<Vec>` instead?

src/core.rs
```diff
@@ -730,20 +990,64 @@ pub fn wrap_optimal_fit<'a, T: Fragment, F: Fn(usize) -> usize>(
...
+            .into_iter()
+            .enumerate()
+            .rev()
+            .map(|(i, data)| {
+                let eol = i + 1 == line_start;
+                if eol {
+                    line_start = minima[line_start].0;
+                }
+                (data.fragment, eol)
+            })
```
mgeisler, line 1006: Have you noticed how this changes the complexity of this piece of code to O(`number_of_fragments`)? Before it was O(`number_of_lines`). Because `wrap_optimal_fit` returns an iterator which yields every single `Fragment`, this seems to be built into the design — which is unfortunate. It feels like a "tax" that we should be able to avoid.

Could `wrap_optimal_fit` and `wrap_first_fit` instead return an `Iterator<Item=Line>` where `Line` is itself something which encapsulates, well, a line of text. This would basically mean grouping the `Fragment`s according to the EOL `bool` you have, but the rest of the layers would no longer deal with individual `Fragment`s but instead deal in full lines. Similar to how the upper layers deal in `&[Fragment]` today.

src/lib.rs
```diff
@@ -652,57 +652,62 @@ where
...
 
+        if broken_words.is_empty() {
+            lines.push(Cow::Borrowed(""));
+            continue;
+        }
+
         #[rustfmt::skip]
         let line_lengths = |i| if i == 0 { initial_width } else { subsequent_width };
+        // This currently isn't as efficient as it could be as we collect into a vec even when it
+        // isn't necessary.
```
mgeisler, line 663: Do you think it will be possible to reconcile this somehow so that both wrapping functions return the same type of iterator? If not, then it seems we lose the efficiency here?

src/lib.rs
```diff
@@ -759,30 +764,21 @@ where
 /// Please see the [`linear`
 /// benchmark](https://github.com/mgeisler/textwrap/blob/master/benches/linear.rs)
 /// for details.
+// TODO: Make this take an &mut str
```
mgeisler, line 767: Please drop from this PR.

mgeisler:
> I think that's the problem here - we have different definitions of "simple" and "complex". I want this API, because I consider iterators to be simpler than vectors, _even if_ they have more lines of code.

Yes, I think you hit the nail on the head there :smile:  I agree that an `Iterator` puts less constraints on the inputs and outputs than a `Vec` or a slice. It's normally great if APIs only makes the minimum assumptions needed — it makes the life easier for the callers.

However, I still think there are a bunch of trade-offs here:

* **Returning `impl Iterator`:** Simpler for the caller: can only call `next()` and that's it. On the other hand, it is less clear for that caller what the _cost_ is of calling `next()` since it's unclear what it does. The two wrapping algorithms illustrate this to some extent: `wrap_first_fit` has an even performance characteristic where every `next()` takes about the same time — `wrap_optimal_fit` has an uneven performance and most of the work is done on the very first call to `next()`. In other words, the `Iterator` API is not the natural API for `wrap_optimal_fit`.

* **Returning `Vec`:** Prescribes memory allocation for the caller. Less choice for the caller which is simpler in some sense, but it might be the wrong choice if the caller did not intend to keep all the wrapped lines in memory at once. The `wrap_first_fit` algorithm can produce streaming output very well, so returning a `Vec` is not the natural API for this function.


> Iterators on the other hand, while they do need more code to implement, are theoretically simpler because they don't do anything other than their singular purpose. So if there's a slight performance loss or there are more types in the public API, so be it. To me it's far better to have a simple API than a fast one or a small one (unless it's _really_ slow of course ).

I get the theoretical purity of this — however, I find it hard to agree that all the `<F as IntoIterator>::Item: Fragment` bounds make the code simpler. It certainly makes it more flexible, but it comes with a cost of extra complexity in my eyes. You can imagine how concerned I was when I introduced the `where Opt: Into<Options<'a, S>>` bounds... it was a nightmare to convince myself :smile: In the end I went with it since I really liked the much smaller API this gave me — no funny `wrap_with_usize` and `wrap_with_options` shims for example.

Similarly, I'm looking for something concrete which is improved by switching things to using `Iterators`. Could you implement a top-level `wrap_iter` function for example. I guess it should be hard-coded to use the `wrap_first_fit` algorithm since this is the one which works well with `Iterator`s.

It would be cool to see a function which can take input in an incremental fashion — that's something which would would require extra refactoring to support today (though the divide-and-conquer approach in #259 seems like it could be generalized). 

> This situation is an interesting clash between the two philosophies of "do one thing and do it well" and using as few lines as possible . I can understand _why_ you'd subscribe to the latter, I just find it a bit reductionist.

One could argue that wrapping text is the goal of the library :smile: Until very recently, this was all it could do: wrap _text_. There was an underlying assumptions that the caller would be able to hold both the input and the output in memory. This makes `Iterator`s mostly uninteresting since they add features that are outside of the core domain of the library.

That being said, #224 makes a case for keeping state while wrapping — I mentioned this above somewhere. So yeah, I would like textwrap to play well with situations where the input isn't know up-front or where the output isn't consumed all at once.

> > textwrap is already showing up as a largeish dependency of clap
> 
> Hmm, this is somewhat of a problem. Maybe more aggressive feature-flagging is the way to go? I see you're already doing that a bit with the other PRs in this repo.

Yeah, I started that recently as a reaction to that discussion. I think it's already pretty modular now... While I don't like the idea of including a huge library for a tiny function, I also don't like the idea that people turn off useful functionality in order to make binaries faster to build — that's what we have caches for. I'm used to Bazel at work which makes it feasible to have both a lot of dependencies plus reasonable fast builds due to extensive caching.

I hope that textwrap can still carry its own weight by providing more correct and better wrapping that what people can hand-roll in an afternoon. The corner case bugs I've fixed over the years tells me that wrapping text isn't completely trivial. The second commit already fixed a small bug: https://github.com/mgeisler/textwrap/commit/0e07230c57ca53c92bf2caa87568afddfbe245ce :smile: 

In any case, we should keep an eye on this aspect too, even if it's a bit annoying.

> > moves three things apart which ought to be one
> 
> Is having them next to each other in the source not enough? I agree it's not quite as together as if it was all in one function, but it's not that much harder to navigate.

Yeah, it's of course not that much harder, but it just feels like a small step backwards to me.  I think it's a matter of ensuring that things are self-contained. If you see `return std::iter::from:fn(...)` then you know that the `Iterator` being returned cannot depend on more than you see right there. It took me a long time to "untangle" the old code and so I'm reluctant to give up on the new structure.

> > I would expect all new libraries to not expose the underlying iterator struct.
> 
> No, the standard library still prefers named types. For example, the recently added `core::future::ready` and `core::future::pending` APIs return `core::future::Ready` and `core::future::Pending`, even when they could be `async fn`s which would the number of LOC. And those are futures, which even I would usually not support writing named types for.

Okay, thanks, I was not aware of this at all :+1:

mgeisler:
To sum up my concerns:

* The performance drop.
* I would like to see the new code do something which cannot be done today.

src/lib.rs
Koxiaet: I want to later introduce an enum iterator that can be `WrapFirstFit` or `WrapOptimalFit`, which will gain back the efficiency hopefully. But I want to keep that in a separate PR.

src/core.rs
Koxiaet: The problem with a `Line` type is that the complexity within the function would still be O(`number_of_fragments`) as each fragment would have to be copied into the line, and it would incur a fairly large allocation cost - each line would probably need a `Vec`.

Almost all code that uses the returned iterator currently iterates over each fragment in the line, so the overall complexity would remain the same most of the time. This approach also avoids double-loops, which _might_ be faster overall, but I'm not sure.

Koxiaet:
> the Iterator API is not the natural API for wrap_optimal_fit.

The natural API for `wrap_optimal_fit` is a reversed iterator. But of course that doesn't exist, so we have to put it into a collection and then reverse it from there. Once we have collected into a vector, we could `.reverse()` it and return that, but we could also use a sequence of `.pop()` calls, and both are equally natural. So my solution was the iterator type, that can both serve as an iterator that calls `.pop()` or can be converted into a `Vec` after calling `.reverse()`. This allows users to use both natural options, so I would consider an iterator the natural API for `wrap_optimal_fit`.

> Could you implement a top-level wrap_iter function for example

I could definitely do this - in fact, this can even be done with the current code. You'd just have to replace the for loop with a `flat_map`. Whether the wrapping functions are iterators doesn't really affect this at all.

> It took me a long time to "untangle" the old code and so I'm reluctant to give up on the new structure.

So I think the solution here is to break up the modules - while it's confusing to have it return a named iterator type that's lower down in the next lines that could be for anything else, it makes it much less confusing if it's all in `wrap_first_fit.rs`. I want to incorporate these ideas in another PR, I have plans for a module structure I think is cleaner than the current one.

> I would like to see the new code do something which cannot be done today.

I should mention the original motivation for this PR. I'm writing some code for a UI library whose API looks roughly like this:

```rust
struct Text { ... }
impl Text {
    fn new(s: String) -> Self { ... }
}
impl Component for Text {
    fn draw(&self, canvas: Canvas) { ... }
}
```

The `new` function will find all the soft line breaks in the text and store their indices (indices because storing many `String`s is very inefficient). Then in `draw` I need to draw all these to the canvas, which can be of any size. With the current API I can't implement `Fragment` for each pair of indices because they don't have access to the string. However, with this new API I can `map` the list of indices, which means they will have access to the string and so I can use a type that implements `Fragment`.

src/core.rs
Koxiaet: Apologies for the aggressive message above, by the way. I was annoyed at the time because this is dragging on so long, but that was no justification for the tone.

I changed my mind - I think that the `FusedIterator` implementation can probably be removed without it harming the user experience at all. Calling `.fuse()` is cheap and barely anyone uses it anyway. So if you do want me to remove it, I would be happy to.

PR #259, opened 2020-12-26
Handle overflows in `wrap_optimal_fit` by divide-and-conquer

### LINKED ISSUE #247: Integer size and overflow when calculating penalty

mgeisler:
The `wrap_optimal_fit algorithm` computes the penalty for a gap as `gap * gap`. If a fragment has a size near `usize::max_value()` and if the line width is small, this computation can easily overflow.

When this happened, we would previously abort or unwind. Now, we instead do the computations with checked arithmetic and detect the overflow. We then proceed to wrap the half of the fragments by themselves. If this work, we then wrap the second half. This way, we might be able to wrap everything without overflow.

Should there be a single fragment which causes the overflow by itself, this fragment is put on a line by itself.

When wrapping part of the fragments, we might of course end up with a partial last line. To fix this, we simply pop this line and re-wrap the fragments that were put onto this line. This ensures no “seams” in the wrapping.

Fixes #247.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -10,7 +10,7 @@ edition = "2018"
 cargo-fuzz = true
 
 [dependencies]
-libfuzzer-sys = "0.3"
+libfuzzer-sys = { version = "0.3", features = ["arbitrary-derive"] }
 textwrap = { path = ".." }
 
 # Prevent this from interfering with workspaces
@@ -26,5 +26,11 @@ doc = false
 [[bin]]
 name = "fill_first_fit"
 path = "fuzz_targets/fill_first_fit.rs"
+test = false
+doc = false
+
+[[bin]]
+name = "wrap_optimal_fit"
+path = "fuzz_targets/wrap_optimal_fit.rs"
 test = false
 doc = false
--- a/fuzz/fuzz_targets/wrap_optimal_fit.rs
+++ b/fuzz/fuzz_targets/wrap_optimal_fit.rs
@@ -1,0 +1,30 @@
+#![no_main]
+use libfuzzer_sys::{arbitrary, fuzz_target};
+use textwrap::core;
+use textwrap::core::Fragment;
+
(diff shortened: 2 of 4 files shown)


mgeisler:
I still need to add a test or two for the new behavior regarding wrapping over-sized fragments. So far I've simply tested that the existing tests still work the same.

mgeisler:
I've added two simple tests for the overflow behavior.

> When wrapping part of the fragments, we might of course end up with a partial last line. To fix this, we simply pop this line and re-wrap the fragments that were put onto this line. This ensures no “seams” in the wrapping.

This is not perfect because I only back up like this in case the current slice resulted in more than _two_ wrapped lines. This is to ensure progress in the wrapping, but it might have the bad side-effect of stitching the wrapped slices together in a bad way.

In general, the above approach implicitly assumes that over-sized fragments are rare so that they occur multiple lines apart. If this is not the case, I don't think the output will be so nice.

mgeisler:
Superseded by #421 which simply makes `wrap_optimal_fit` return a `Result` instead of trying to do to something fancy and handle overflows by recursion.

PR #279, opened 2021-01-06
Make indent preserve the of \n in the input string (stop adding a trailing \n) [stable rustc]

### LINKED ISSUE #207: The indent function always adds a final \n, even if the input string has no \n

jRimbault:
I ported a custom iterator to get the inclusive split feature specifically with a `char` (though it could be even be just a custom iterator with `\n` as a const value).

This fixes #207 on stable rust but I'd feel safer once we're able to use the much more generic `split_inclusive`, hopefully.

~It uses `unsafe` code. Which is forbidden in this crate.~ It's up to you @mgeisler if you want to make an exception or not.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,7 @@ env:
 jobs:
   build:
     strategy:
+      fail-fast: false
       matrix:
         os:
           - ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,46 +10,67 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  build:
+
+  ubuntu:
+    name: Build on Ubuntu
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        os:
-          - ubuntu-latest
-          - windows-latest
         rust:
           - stable
           - nightly
 
-    runs-on: ${{ matrix.os }}
-
     steps:
-      # By default, the "\n" in CHANGELOG.md end up as "\r\n" on disk.
-      # This breaks the version-numbers test since ^...$ no longer
-      # match because the regex crate doesn't treat \r\n as a line
-      # boundary (https://github.com/rust-lang/regex/issues/244).
-      - name: Configure Git line ending handling
-        run: git config --global core.autocrlf input
-
       - name: Checkout repository
         uses: actions/checkout@v2
 
       - name: Install ${{ matrix.rust }} Rust
         run: rustup default ${{ matrix.rust }}
 
-      # This gives us a quick smoke test in case something is
-      # seriously broken. It also gives us valuable information about
-      # the typical compile time.
-      - name: Build with default features
-        run: cargo build
+      - name: Test with default features
+        run: cargo test
 
-      - name: Build with all features
-        run: cargo build --all-features
+      - name: Test without default features
+        run: cargo test --no-default-features
 
-      # This builds examples and benchmarks. Because benchmarks are a
-      # nightly-only feature, we skip this on stable Rust.
-      - name: Build all targets with all features
-        if: matrix.rust == 'nightly'
-        run: cargo build --all-targets --all-features --benches
-
-      - name: Run tests
+      - name: Test with all features
         run: cargo test --all-features
+
+  windows:
+    name: Build on Windows (stable)
+    runs-on: windows-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Test with default features
+        run: cargo test
+
+  # This builds examples and benchmarks, which are not covered above.
+  build-all:
+    name: Build all targets
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Build all targets with default features
+        run: cargo build --all-targets
+
+      - name: Build all targets without default features
+        run: cargo build --all-targets --no-default-features
+
+      - name: Build all targets with all features
+        run: cargo build --all-targets --all-features
+
+  build-documentation:
+    name: Build documentation
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Build documentation and check intra-doc links
+        env:
+          RUSTDOCFLAGS: --deny broken_intra_doc_links
+        run: cargo doc --all-features --no-deps
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,6 @@ env:
 jobs:
   build:
     strategy:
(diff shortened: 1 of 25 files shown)


mgeisler:
> ~It uses `unsafe` code. Which is forbidden in this crate.~ It's up to you @mgeisler if you want to make an exception or not.

Hey @jRimbault, thanks for removing the unsafe code again! Until someone shows up with a performance problem (with benchmarks...), there's no need for unsafe code here :smile:

.github/workflows/build.yml
```diff
@@ -12,6 +12,7 @@ env:
 jobs:
   build:
     strategy:
+      fail-fast: false
```
mgeisler, line 15: If we need this, then please move it to a separate PR. In general, I've been okay happy with failing PRs on the first failure since it normally indicates a real error that the PR author should fix.

src/indentation.rs
```diff
@@ -45,20 +44,61 @@
...
 ///
-/// assert_eq!(indent(" \t  Foo   ", "->"), "-> \t  Foo   \n");
+/// assert_eq!(indent(" \t  Foo   ", "->"), "-> \t  Foo   ");
 /// ```
 pub fn indent(s: &str, prefix: &str) -> String {
     let mut result = String::new();
-    for line in s.lines() {
-        if line.chars().any(|c| !c.is_whitespace()) {
+
+    for line in CharSplitInclusive::new(s, '\n') {
```
mgeisler, line 52: Could this be simplified by instead using
```rust
for idx, line in s.split('\n').enumerate() {
    // ... normal logic
    if idx > 0 {
        result.push('\n');
    }
}

I _think_ this should add exactly the name number of `\n` characters as we found originally.

File: tests/indent.rs [hunk 3]
```diff
@@ -0,0 +1,88 @@
...
+    for text in ROUNDTRIP_CASES.iter() {
+        assert_eq!(&dedent(&indent(text, " \t  \t ")), text);
+    }
+}
+
+#[test]
+fn test_indent_default() {
+    // Test default indenting of lines that are not whitespace only
+    let prefix = "  ";
+    let expected = [
```
mgeisler, line 59: Is this testing something which isn't already covered in the unit tests in `src/indentations.rs`?

tests/indent.rs
```diff
@@ -0,0 +1,88 @@
...
+    for text in ROUNDTRIP_CASES.iter() {
+        assert_eq!(&indent(text, ""), text);
+    }
+    for text in WINDOWS_CASES.iter() {
+        assert_eq!(&indent(text, ""), text);
+    }
+}
+
+#[test]
+fn test_roundtrip_spaces() {
```
mgeisler, line 32: Could you instead add a fuzz test for this? That would be much more powerful since it would explore many more test cases.

Please see #256 for two existing examples. You'll need [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz) installed and then it's pretty easy to add a new fuzz test with `cargo fuzz add indent`.

tests/indent.rs
```diff
@@ -0,0 +1,88 @@
...
+        // Pathological case
+        "  Hi.\r\n  This is a test.\n\r\n  Testing.\r\n\n",
+    ];
+    for (text, expect) in WINDOWS_CASES.iter().zip(expected.iter()) {
+        assert_eq!(&indent(text, prefix), expect)
+    }
+}
+
+#[test]
+fn indented_text_should_have_the_same_number_of_lines_as_the_original_text() {
```
mgeisler, line 82: This could also be expressed very elegantly as a fuzz test.

jRimbault:
I'll get on to it soon :)

mgeisler:
Hi @jRimbault, no stress! That's the beauty of open source :smile:

mgeisler:
... and thanks!

.github/workflows/build.yml
jRimbault: Of course, this was only intended for this branch, thanks for the reminder.

src/indentation.rs
jRimbault: That's actually much more clever than everything I did before ^^

And it does pass the tests.

tests/indent.rs
jRimbault: I ported this whole test suite (file) from the tests of the python `textwrap` package, I actually don't remember the corner cases it tested, but I trust it ? :/

jRimbault:
I tried making some fuzz work but I don't understand quite what it's testing for ? Panics ? The examples here don't seem to actually test things ?

In any case I think your solution to simply add an enumerating index is really quite good since it doesn't need bumping the MSRV (`split_inclusive` will be part of 1.51). I also think, rather than fuzzing, taking tests from tried and proved libraries like, this crate inspiration, python's `textwrap` should be a better faster option.

mgeisler:
> I tried making some fuzz work but I don't understand quite what it's testing for ? Panics ? The examples here don't seem to actually test things ?

Yes, you raise a good point and you guessed correctly: they simply test that the code doesn't panic :smile: I should really document this...

> In any case I think your solution to simply add an enumerating index is really quite good since it doesn't need bumping the MSRV (`split_inclusive` will be part of 1.51). I also think, rather than fuzzing, taking tests from tried and proved libraries like, this crate inspiration, python's `textwrap` should be a better faster option (CPU time, and reliability wise).

Fair enough, I like it a lot that you ported those tests. They give us a nice foundation and we can then always think about turning them into property-based tests later. You're definitely right that it's much cheaper to have some solid unit tests for this.

tests/indent.rs
mgeisler: Okay, let's go with this then!

mgeisler (approved):

jRimbault:
Happy to help 😄

PR #289, opened 2021-02-07
Switch to `f64` for computing penalties in `wrap_optimal_fit`

mgeisler:


mgeisler:
Superseded by #421 which changes the `Fragment` trait to use `f64`. This PR changes the cost computation to a different style which normalizes the cost between zero and a given maximum. This prevents overflows, but it also means makes it harder to reason about what the penalties mean. The change in #421 keeps the existing meaning of the cost computation unchanged and seems more flexible because it makes fragments more flexible in general.

PR #310, opened 2021-04-05
Add WebAssembly example

mgeisler:
This example shows how one can use Textwrap to wrap text outside of the console. In this case, the text is shown on a HTML canvas. The user can switch the font family between monospace, serif, and sans-serif and we accurately use the actual width as measured by the browser to wrap the text.

The code is manually deployed to https://mgeisler.github.io/textwrap/, I'll work on some automation for this next.


SirWindfield:
Not sure if you are aware of it, but the example doesn't work in Firefox. Not a huge deal, but might be good to add a notice somewhere:
image

mgeisler:
Hi @SirWindfield, thanks for testing it! I would love to make it work in Firefox too and you're right, it should be noted somewhere what I've tested it with.

mgeisler:
Hi @SirWindfield,

I tested it some more and found the problem — I was using some methods which aren't yet supported by Firefox. A new version has been deployed in 9aaa7c50221bd8388733655a98b48bcf5ed9e932.

Issue #220, opened 2020-11-08
Does not work for languages without word separators


Kestrer:
For example, `fill("កើតមកមានសេរីភាព", 6)` outputs `កើតមកម`, `ានសេរភ` and `ាព` when it should output `កើតមក`, `មាន` and `សេរីភាព`. See also [w3's Approaches to line breaking document](https://www.w3.org/International/articles/typography/linebreak.en#sec_se_asia) which has the correct ways to line break words; implementing support for this would require storing a dictionary and matching words in it.

mgeisler:
Hi @Koxiaet, thanks you very much for linking me to the w3 article. I had not read that before and it's very interesting to get some background information on this.

You're completely right that `textwrap` only breaks on whitespace right now — I would like to extend this, as mentioned in #80 about how East-Asian languages could be handled.

I'm currently working on reimplementing the basic line breaking algorithm. The goal is to make it more robust and flexible. Right now, the algorithm naively goes through the string and breaks the string into words based on whitespace. The processing is trying to do everything at once and it's not easy to adjust or extend it to handle more complicated cases like you describe.

Do you know if there are freely available dictionaries we can refer to here? I know Danish, English and German myself — all written the same way with Latin script... So I'm very much out of my depth here and would need all the help I can get!

Koxiaet:
I'm also out of my depth here, I don't know any non-European languages either. I think the best option is to use [`rust_icu_ubrk`](https://crates.io/crates/rust_icu_ubrk) which provides Rust wrappers for [Unicode's boundary analysis algorithms in its ICU library](https://unicode-org.github.io/icu/userguide/boundaryanalysis/).

mgeisler:
When I was looking into this previously, I was considering to use [`unicode-linebreak`](https://crates.io/crates/unicode-linebreak), but I think this crate only does a subset of what you're talking about: I think it only finds breakpoints around spaces and hyphens, and it forbids some breakpoints around certain kinds of punctuation.

An an example, it will find no breakpoints in `"Hello !"` since the space before the `!` doesn't count for some reason specified in [Unicode Standard Annex #14](https://www.unicode.org/reports/tr14/).

I guess the breaks computed by `rust_icu_ubrk` are quite different if they can also occur inside words?

I'll probably not make a lot of progress on this issue by myself since I'm not an expert in this field. I'll leave it open in case such an expert comes by :-)

Koxiaet:
[...]

Issue #80, opened 2017-08-26
Investigate how textwrap works for East Asian languages


mgeisler:
I wonder if it would be possible to tweak the wrapping so as to follow some of the rules mentioned here:
https://en.wikipedia.org/wiki/Line_breaking_rules_in_East_Asian_languages

mgeisler:
I'll close the issue now: the new `WordSeparator` trait makes it possible to customize exactly how a line of text is slices up into "words" or "syllables" or "Ideographs".

The `UnicodeBreakProperties` word separator merged in #313 uses the Unicode line breaking algorithm to find words and will handle CJK characters according to the Unicode rules. This means that we can find words without intervening whitespace:

```rust
assert_eq!(UnicodeBreakProperties.find_words("CJK: 你好").collect::<Vec<_>>(),
           vec![Word::from("CJK: "),
                Word::from("你"),
                Word::from("好")]);
```

These words are then wrapped by measuring their width and putting them into lines as usual.

If someone with experience in these languages finds problems, then I would love to hear about them since I'm not an expert at all in these things.

Issue #14, opened 2017-01-16
Avoid splitting on initial hyphens in --foo-bar


mgeisler:
We should only split on hyphens surrounded by non-hyphens or perhaps surrounded by word-characters.

PR #313, opened 2021-04-08
Use Unicode line breaking algorithm to find words

mgeisler:
This adds a new optional dependency on the [unicode-linebreak crate](https://docs.rs/unicode-linebreak) which implements the line breaking algorithm from [Unicode Standard Annex #14](https://www.unicode.org/reports/tr14/).

The new dependency is enabled by default since these line breaks are more correct than what you get by splitting on whitespace.

This should help address #220 and #80, though I’m no expert on non-Western languages. More feedback from the community would be needed here.


src/core.rs
```diff
@@ -326,19 +343,82 @@ impl Fragment for Word<'_> {
...
+/// assert_eq!(find_words("[ foo ] bar !", LineBreakAlgorithm::UnicodeLineBreaks),
+///            vec![Word::from("[ foo ] "),
+///                 Word::from("bar !")]);
+/// ```
+pub fn find_words(line: &str, line_break_algorithm: LineBreakAlgorithm) -> Vec<Word> {
+    match line_break_algorithm {
+        LineBreakAlgorithm::Whitespace => find_ascii_words(line).collect::<Vec<_>>(),
+        #[cfg(feature = "unicode-linebreak")]
+        LineBreakAlgorithm::UnicodeLineBreaks => find_unicode_words(line).collect::<Vec<_>>(),
+    }
```
mgeisler, line 417: This needs to go — I'll introduce a trait instead so we can avoid branches at runtime and instead push them to compile time.

src/core.rs
```diff
@@ -371,6 +451,86 @@ pub fn find_words(line: &str) -> impl Iterator<Item = Word> {
...
+        result
+    }
+
+    // Construct an iterator over (original index, stripped index)
+    // tuples. We find the Unicode linebreaks on a stripped string,
+    // but we need the original indices so we can form words based on
+    // the original string.
+    let mut last_stripped_idx = 0;
+    let mut char_indices = line.char_indices();
+    let mut idx_map = std::iter::from_fn(move || match char_indices.next() {
```
mgeisler, line 478: This is all terribly complex and somewhat slow — I'll rework it to have a fast path for the case where we don't see any `CSI` characters in the string.

src/core.rs
SirWindfield: I just wanted to comment on this :) Yep, I do also think that a trait is a better idea than an enum here. It just introduces something that the maintainers have to keep track of and change if someone wants their algorithm to get included/be accessible to `textwrap`.

mgeisler: Yes, I agree. Mostly because of how a trait will streamline the code. The openness of a trait is a secondary benefit: so far, I don't think anyone has made their own implementation of the `WordSplitter` trait.

mgeisler: I've updated the PR with a new `LineBreakAlgorithm` trait. I'm not super happy with the names here, so I might change my mind in the next few days...

src/line_breaks.rs
```diff
@@ -0,0 +1,367 @@
...
+/// to handle hyphenation of individual words.
+pub trait LineBreakAlgorithm: std::clone::Clone {
+    /// Find line break opportunities in `line` and return these as words.
+    fn break_line<'a>(&self, line: &'a str) -> Box<dyn Iterator<Item = Word<'a>> + 'a>;
+}
+
+/// Default line breaking algorithm when the `unicode-linebreak` Cargo
+/// feature is enabled.
+#[cfg(feature = "unicode-linebreak")]
+pub type DefaultLineBreakAlgorithm = UnicodeLineBreaks;
```
mgeisler, line 25: This huge identifier is annoying to see and it creates an annoying indirection... it's there since I feel the unicode-linebreak crate is an optional dependency, yet I would like to use it by default if it is available. So to avoid tons of `#[cfg(feature = "unicode-linebreak")]` in the rest of the code, I define this type here.

mgeisler:
I'll need to make a small detour before I can continue with this PR — the `?Sized` bound for the `WordSplitter` trait cannot be generalized since we can only have one unsized field in a struct. I'll rework this first and then we can continue with this PR.

src/core.rs
mgeisler: Will do this in #333.

src/line_breaks.rs
mgeisler: It turns out I can use a macro for this! This allows me to keep the shenanigans here internal to the library.

mgeisler:
> I'll need to make a small detour before I can continue with this PR — the `?Sized` bound for the `WordSplitter` trait cannot be generalized since we can only have one unsized field in a struct. I'll rework this first and then we can continue with this PR.

The detour has been completed with #331 and #328.

src/core.rs
mgeisler: I changed my mind and we now have a `WordSeparator` trait. I've also reworked the existing `WordSplitter` trait so that that the two traits work in a consistent way, see #331 and #332.

PR #352, opened 2021-05-08
Combined struct Word's whitespace and penalty fields to PostFix enum

NeverGivinUp:
Replacing the fields with the enum. This is minimal invasive. The tests are unchanged as far as possible. Notice, how I had to make several distinctions in `Word::from` and `Word::break_apart` to get the tests working. 

It remains confusing for me, when an empty whitespace and when an empty penalty are appropriate...

Also, I wasn't able to run the benchmark. Cargo Criterion breaks down for me with https://github.com/bheisler/criterion.rs/issues/466

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -49,7 +49,7 @@ const ANSI_FINAL_BYTE: std::ops::RangeInclusive<char> = '\x40'..='\x7e';
 /// `chars` provide the following characters. The `chars` will be
 /// modified if `ch` is the start of an ANSI escape sequence.
 #[inline]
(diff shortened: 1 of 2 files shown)


mgeisler:
Hey Navid, thanks a lot!

I won't be able to review and merge this today, but I have an initial comment: there are some spurious whitespace changes — perhaps one of us is using an old `rustfmt`? :-)

It would be great if you could revert those changes so the diff only contains your actual changes. Thanks!

mgeisler:
Hi Navid,
I'm sorry this didn't work out — it sounded like a nice idea at first, but seeing it in action makes me think that it doesn't improve the complexity like we hoped.

PR #377, opened 2021-06-03
Update `hyphenation` to v0.8.2

tapeinosyne:
I observe one test failing on my machine, but it does so with `hyphenation` disabled as well:

```
thread 'core::tests::display_width_emojis' panicked at 'assertion failed: `(left == right)`
  left: `18`,
 right: `20`', src/core.rs:432:9
```

Perhaps more importantly, CI fails on `deploy WASM demo`: https://github.com/tapeinosyne/textwrap/runs/2740901118?check_suite_focus=true

I am unfamiliar with how textwrap compiles to WASM, which means I have little to contribute on the issue. However, do let me know if further changes are needed on my end.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,8 +35,7 @@ unicode-linebreak = { version = "0.1", optional = true }
 unicode-width = { version= "0.1", optional = true }
 
 [dependencies.hyphenation]
-git = "https://github.com/tapeinosyne/hyphenation"
-rev = "d8d501a3731d"  # Until `Standard` implements `Clone`
+version = "0.8.2"
 optional = true
 features = ["embed_en-us"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,15 +29,11 @@ path = "benches/indent.rs"
 default = ["unicode-linebreak", "unicode-width", "smawk"]
 
 [dependencies]
+hyphenation = { version = "0.8.2", optional = true, features = ["embed_en-us"] }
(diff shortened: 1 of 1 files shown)


mgeisler:
Wow, thanks for much for the release and the PR!

> I observe one test failing on my machine, but it does so with `hyphenation` disabled as well:
> 
> ```
> thread 'core::tests::display_width_emojis' panicked at 'assertion failed: `(left == right)`
>   left: `18`,
>  right: `20`', src/core.rs:432:9
> ```

That error is actually strange to me... determining if a `char` is an emoji should be completely deterministic. I see that the tests pass in CI when run in your clone, and they also pass here.

The test checks that 10 emojis have a width of 20 columns:
```rust
assert_eq!(display_width("😂😭🥺🤣✨😍🙏🥰😊🔥"), 20)
```

This is unrelated to the `hyphenation` feature as you say, so I'll just merge this for now.

> Perhaps more importantly, CI fails on `deploy WASM demo`: https://github.com/tapeinosyne/textwrap/runs/2740901118?check_suite_focus=true
> 
> I am unfamiliar with how textwrap compiles to WASM, which means I have little to contribute on the issue. However, do let me know if further changes are needed on my end

Thanks for letting me know! That step tries to checkout the `gh-pages` branch from this repository — and I think this simply fails because your clone doesn't seem to have this branch: https://github.com/tapeinosyne/textwrap/branches/all. So that error is fairly innocent.

PR #381, opened 2021-06-05
Update `hyphenation` to v0.8.3

tapeinosyne:
No functional changes for `textwrap`: the English dictionary is now built with an updated version of `bincode`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ path = "benches/indent.rs"
 default = ["unicode-linebreak", "unicode-width", "smawk"]
 
 [dependencies]
-hyphenation = { version = "0.8.2", optional = true, features = ["embed_en-us"] }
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @tapeinosyne,
If it's okay with you, I would like to leave the dependency at 0.8.2. Cargo will automatically use 0.8.3 when users run `cargo update` and since we don't _need_ the 0.8.3 version, the less restrictive 0.8.2 bound is preferable.

When you release a 0.9.0 some day, then I'll be happy to bump the dependency to say `version = "0.9"`, so that users get the benefits of the new 0.9.x versions.

tapeinosyne:
> If it's okay with you, I would like to leave the dependency at 0.8.2.

Of course; there are no compatibility issues.

I'll let you know when v0.9 is released.

PR #389, opened 2021-06-22
Introduce parameters to `OptimalFit`

### LINKED ISSUE #326: Make optimal-fit penalties adjustable
The current penalties work well for terminal text, but the fixed penalties don't work for text outside of the terminal. An an example, the
```rust
/// Penalty for lines ending with a hyphen.
const HYPHEN_PENALTY: i32 = 25;
```
penalty basically says "it costs the same to insert a hyphen as leaving a gap of 5 columns at the end of the line". This no longer makes sense if the font is variable width and the magnitude of the penalty doesn't match if we start measuring line lengths in pt or px.

mgeisler:
This turns the hard-coded penalty constants used by `wrap_optimal_fit` into adjustable parameters. This makes it feasible to wrap proportional text using the optimal-fit algorithm.

Fixes #326.


src/wrap_algorithms/optimal_fit.rs
```diff
@@ -6,18 +6,107 @@ use crate::wrap_algorithms::WrapAlgorithm;
...
+    /// ```
+    ///
+    /// This only happens if the overflowing word is 50 characters
+    /// long _and_ if the word overflows the line by exactly one
+    /// character. If it overflows by more than one character, the
+    /// overflow penalty will quickly outgrow the cost of the gap, as
+    /// seen above.
+    pub overflow_penalty: i32,
+
+    /// When should the last line be considered "short".
```
robinkrahl, line 72: Could you maybe add some more information about the meaning of this value?  I understand that *width* <= *max_width* / *short_line_fraction* is considered a short line, but the description is not entirely clear.

mgeisler: Yeah... great point, thanks! I've expanded it a bit now. Does it look better?

An example would of course be great, but I wasn't able to quickly conjure one up with the Wasm demo.

robinkrahl: Much better, thanks!

mgeisler: Thanks, much appreciated!

Issue #390, opened 2021-06-25
Panic due to byte access between char boundary


MakotoE:
When you run this unit test, it panics with `byte index 3 is not a char boundary`.

```rust
    #[test]
    fn test() {
        let s = "\u{1b}!Ͽ";
        fill(s, 10);
    }
```

MakotoE:
I'm just guessing here, but was this piece of code supposed to be like:

```diff
diff --git a/src/word_separators.rs b/src/word_separators.rs
index 81ec6d4..abac756 100644
--- a/src/word_separators.rs
+++ b/src/word_separators.rs
@@ -216,7 +216,7 @@ impl WordSeparator for UnicodeBreakProperties {
         let mut opportunities = unicode_linebreak::linebreaks(&stripped)
             .filter(|(idx, _)| {
                 #[allow(clippy::match_like_matches_macro)]
-                match &line[..*idx].chars().next_back() {
+                match &stripped[..*idx].chars().next_back() {
                     // We suppress breaks at ‘-’ since we want to control
                     // this via the WordSplitter.
                     Some('-') => false,
```

mgeisler:
Hi @MakotoE, thank you very much for the report! I think you're right — it seems very suspicious that we iterate over `stripped` but slice `line`.

I'm not at a computer right now, but I assume your test no longer fails if you swap the variables? If so, could you open a PR? Thanks!

Issue #280, opened 2021-01-21
Integrate fuzz tests in CI


mgeisler:
We could run the [fuzz tests](https://github.com/mgeisler/textwrap/tree/master/fuzz/fuzz_targets) for, say, 10 seconds on every PR. It looks like the command to run is

```sh
$ cargo fuzz run fill_first_fit -- -max_total_time=10
```

We would need to figure out how to run all the fuzz targets, preferably in parallel.

PR #391, opened 2021-06-25
Fix panic in `find_words` due to string access outside of a character boundary

MakotoE:
Closes https://github.com/mgeisler/textwrap/issues/390

This fixes an attempt to access string bytes that may be within a character boundary.

Please check if this is actually how the code should be. All I know is that all tests pass.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1852,6 +1852,12 @@ mod tests {
     }
 
     #[test]
+    fn fill_unicode_boundary() {
+        // https://github.com/mgeisler/textwrap/issues/390
+        fill("\u{1b}!Ͽ", 10);
+    }
+
+    #[test]
     #[cfg(not(feature = "smawk"))]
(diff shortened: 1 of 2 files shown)


mgeisler:
Thanks! I'll double-check this when I'm home and then merge. We should really address #280 since I hope it could have cought this.

MakotoE:
> We should really address #280 since I hope it could have cought this.

I wouldn't run fuzz tests for all CI runs because you want commit tests to be fast. I think it should be run every night (on days when commits are merged).

mgeisler:
Thanks for the PR, I'll make a 0.14.1 release with the fix now.

> I wouldn't run fuzz tests for all CI runs because you want commit tests to be fast.

Yeah, I wouldn't want to prolong the tests by half or full hours. in this case, it turns out that the bug can be found by fuzzing for less than a minute. So I'll setup very brief fuzz testing on PRs, just to have a quick sanity check in the future.

PR #392, opened 2021-06-25
wrap_optimal_fit() - Checked arithmetic

MakotoE:
This fixes an arithmetic overflow due to passing a large width and a certain string. I did not replace all the possible overflow operations in `wrap_optimal_fit()` because they did not cause a crash during fuzz testing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1852,6 +1852,11 @@ mod tests {
     }
 
     #[test]
+    fn fill_large_width() {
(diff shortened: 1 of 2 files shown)


mgeisler:
Hey @MakotoE, the bug you're fixing is discussed in #247. I've been looking at different ways of fixing it — please see the other open PRs.

mgeisler:
Fixed by #421.

PR #398, opened 2021-06-27
Cache compilation artifacts with sccache

mgeisler:


mgeisler:
It seems easier to cache the Cargo directories like in #399.

PR #400, opened 2021-06-27
add fill_optimal_fit fuzz target to CI

MakotoE:
I noticed that the `fill_optimal_fit` fuzz target was not run in CI so this PR makes the build workflow run this target right after `fill_first_fit`. `fill_first_fit` alone does not catch [this issue](https://github.com/mgeisler/textwrap/pull/392).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -132,13 +132,19 @@ jobs:
             fuzz-corpus-
 
       - name: Build fuzz targets
-        run: cargo fuzz build fill_first_fit
+        run: |
+          cargo fuzz build fill_first_fit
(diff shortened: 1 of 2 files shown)


MakotoE:
Good, the fuzz test caught the bug.

> `thread '<unnamed>' panicked at 'attempt to multiply with overflow', /home/runner/work/textwrap/textwrap/src/wrap_algorithms/optimal_fit.rs:286:21`

mgeisler:
Yeah, the `fill_optimal_fit` test fails, that's why I didn't add it to CI yet. It's #247 again :-)

mgeisler:
This is now fixed via #421.

PR #402, opened 2021-07-01
Make word separators and splitters more flexible

robinkrahl:
This PR makes the word separators and splitters more flexible to allow users to use their own words types.  It also adds an example that shows how to use textwrap with custom word types, namely with styled strings.

As this is only a draft, I did not add much documentation.

The basic changes are:

- For word separation, introduce a new method that just returns the range of the detected words.  Users can then apply their own logic for creating words from these ranges.
- For word splitting, introduce a new `Fragments` iterator struct that yields the fragments for a word.  It turned out to be easier to introduce a `Splittable` trait than to use a closure to perform the splitting, but this could also be changed.  I used a new struct for the iterator because that makes it much easier to keep track of the life times.

### DIFF
--- a/examples/style.rs
+++ b/examples/style.rs
@@ -1,0 +1,134 @@
+use hyphenation::{Language, Load, Standard};
+use rand::Rng as _;
+use textwrap::word_separators::WordSeparator as _;
+
+#[derive(Debug)]
+struct StyledWord<'a> {
+    word: &'a str,
+    whitespace: &'a str,
+    hyphen: bool,
+    style: Option<text_style::Style>,
+}
+
+impl StyledWord<'_> {
+    fn render(&self, is_end: bool) {
+        use text_style::termion::Termion as _;
+
+        print!(
+            "{}",
+            text_style::StyledStr::new(self.word, self.style).termion()
+        );
+
+        if is_end {
+            if self.hyphen {
+                print!("{}", text_style::StyledStr::new("-", self.style).termion());
+            }
+        } else {
+            print!("{}", self.whitespace);
+        }
+    }
+}
+
+impl AsRef<str> for StyledWord<'_> {
+    fn as_ref(&self) -> &str {
+        &self.word
+    }
+}
+
+impl<'a> From<text_style::StyledStr<'a>> for StyledWord<'a> {
+    fn from(word: text_style::StyledStr<'a>) -> Self {
+        let trimmed = word.s.trim_end_matches(' ');
+        Self {
+            word: trimmed,
+            whitespace: &word.s[trimmed.len()..],
+            hyphen: false,
+            style: word.style,
+        }
+    }
+}
+
+impl textwrap::core::Fragment for StyledWord<'_> {
+    fn width(&self) -> usize {
+        self.word.len()
+    }
+
+    fn whitespace_width(&self) -> usize {
+        self.whitespace.len()
+    }
+
+    fn penalty_width(&self) -> usize {
+        if self.hyphen {
+            1
+        } else {
+            0
+        }
+    }
+}
+
+impl textwrap::word_splitters::Splittable for StyledWord<'_> {
+    type Output = Self;
+
+    fn split(&self, range: std::ops::Range<usize>, keep_ending: bool) -> Self::Output {
+        let word = &self.word[range];
+        Self {
+            word,
+            whitespace: if keep_ending { self.whitespace } else { "" },
+            hyphen: if keep_ending {
+                self.hyphen
+            } else {
+                !word.ends_with('-')
+            },
+            style: self.style,
+        }
+    }
+}
+
+fn generate_style(rng: &mut impl rand::Rng) -> text_style::Style {
+    let mut style = text_style::Style::default();
+
+    style.set_bold(rng.gen_bool(0.1));
+    style.set_italic(rng.gen_bool(0.1));
+    style.set_underline(rng.gen_bool(0.1));
+    style.strikethrough(rng.gen_bool(0.01));
+
+    style.fg = match rng.gen_range(0..100) {
+        0..=10 => Some(text_style::AnsiColor::Red),
+        11..=20 => Some(text_style::AnsiColor::Green),
+        21..=30 => Some(text_style::AnsiColor::Blue),
+        _ => None,
+    }
+    .map(|color| text_style::Color::Ansi {
+        color,
+        mode: text_style::AnsiMode::Light,
+    });
+
+    style
+}
+
+fn main() {
+    let dictionary = Standard::from_embedded(Language::EnglishUS).unwrap();
+    let mut rng = rand::thread_rng();
+
+    let text = lipsum::lipsum(rng.gen_range(100..500));
+
+    let styled = text
+        .split_inclusive(' ')
+        .map(|s| text_style::StyledStr::styled(s, generate_style(&mut rng)));
+    let words: Vec<_> = styled
+        .flat_map(|s| {
+            textwrap::word_separators::AsciiSpace
+                .find_word_ranges(&s.s)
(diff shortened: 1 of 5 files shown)


src/word_splitters.rs
```diff
@@ -190,36 +190,79 @@ where
...
+            },
+        }
+    }
+}
+
+#[allow(missing_docs)]
+#[derive(Debug)]
+pub struct Fragments<W: Splittable, I: Iterator<Item = usize>> {
+    word: W,
+    split_points: I,
```
robinkrahl, line 230: This type parameter is unnecessary, we can directly use `std::vec::IntoIter<usize>`.

mgeisler:
Hey @robinkrahl, I just wanted to say that I think this looks really nice!

One thought: would it make sense to combine `Splittable` with `Fragment` and thus say that any type which a `Fragment` should have a way of splitting itself? The trait could have a default implementation which doesn't do any splitting.

So far, my thinking has been that fragments contain just the necessary information for us to be able to wrap them nicely with the two wrapping algorithms. That is, a fragment has a width and whitespce/penalty following it. However, since the proposed wrapping "pipeline" also involves finding and splitting the fragments (`Word`s), perhaps it would be much nicer to combine the whole thing into one.

Just a quick idea... I'll give it a proper look tomorrow.

robinkrahl:
Thanks for the update!

> One thought: would it make sense to combine `Splittable` with
> `Fragment` and thus say that any type which a `Fragment` should have a
> way of splitting itself? The trait could have a default implementation
> which doesn't do any splitting.

Good point.  I had a similar thought.  The reason why I didn’t want to
combine these two is that the width computation is not required for the
word splitting, it might be expensive and might require more information
than just the string and the style.  Therefore, users might want to have
separate types for a splittable word and a fragment with width
calculations that can also be cached.

(Even the current implementation for core::Word currently performs
unnecessary width calculations for words that are split up later, though
the cost is probably negligible.)

In my use case, I have to look up the glyphs in the font data to
calculate their widths.  Therefore, my pipeline currently is:  separate
and split words using string slices (→ StyledWord), calculate their
widths and copy the strings (→ StyledFragment), cache the result and
then run the wrap algorithm.

Of course this could also be implemented with a single struct, but it
would have to keep some state.

These are no big issues, so if you prefer a single trait, I could live
with that too.

src/word_splitters.rs
```diff
@@ -190,36 +190,71 @@ where
...
+            word,
+            width: display_width(word),
+            whitespace: if keep_ending {
+                self.word.whitespace
+            } else {
+                ""
+            },
+            penalty: if keep_ending {
+                self.word.penalty
+            } else if word.ends_with('-') {
```
robinkrahl: This condition should be negated – fixed in the next commit.

mgeisler:
> Therefore, users might want to have separate types for a splittable word and a fragment with width calculations that can also be cached.

Yes, I can see why that could be useful for users of the library. On the other hand, I wrote a script to download every [reverse dependency](https://crates.io/crates/textwrap/reverse_dependencies), and I cannot find any public usages of `textwrap::core`, except for Clap which uses ` textwrap::core::display_width`. The more advanced functions seems unused except by my own examples — and perhaps genpdf at some point :laughing:

Basically, we seem to have pretty free hands in how we structure the internals, while still making a 0.15.0 release a drop-in replacement for everybody. If two traits makes your life easier, than I'm happy to support it.

> (Even the current implementation for core::Word currently performs unnecessary width calculations for words that are split up later, though the cost is probably negligible.)

You're right and I remember toying with the idea of having an `UnmeasuredWord` struct back in the day. In the end I figured it was enough with `Word` and then measure the text twice. The difference in performance is about 7% on a quick test I just did — but given that it takes around 260 microseconds to wrap a 4800 character long line on my laptop, I think the performance is good enough, even for realtime wrapping in a GUI or similar.

> These are no big issues, so if you prefer a single trait, I could live with that too.

No, let's go with two since it seems to better model the information necessary at different stages. Would it perhaps make sense to have

* `SplittableFragment`: an abstract piece of text from which you can get a sub-fragment (your `Splittable` trait)
* `MeasuredFragment`: an abstract piece of text which known how big it is and which we can word wrap (the existing `Fragment` trait)

I'm thinking names like that might help to tie the two concepts together, while also being more descriptive. Please let me know what you think!

robinkrahl:
>     * `SplittableFragment`: an abstract piece of text from which you can get a sub-fragment (your `Splittable` trait)
> 
>     * `MeasuredFragment`: an abstract piece of text which known how big it is and which we can word wrap (the existing `Fragment` trait)
> 
> 
> I'm thinking names like that might help to tie the two concepts together, while also being more descriptive. Please let me know what you think!

Sounds good!

mgeisler:
> > * `SplittableFragment`: an abstract piece of text from which you can get a sub-fragment (your `Splittable` trait)
> > * `MeasuredFragment`: an abstract piece of text which known how big it is and which we can word wrap (the existing `Fragment` trait)
> > 
> > I'm thinking names like that might help to tie the two concepts together, while also being more descriptive. Please let me know what you think!
> 
> Sounds good!

Okay, cool — let's do that independently of this PR.

src/word_splitters.rs
```diff
@@ -190,36 +190,79 @@ where
...
+        if let Some(idx) = self.split_points.next() {
+            let w = self.word.split(self.prev..idx, false);
+            self.prev = idx;
+            return Some(w);
+        }
+
+        let len = self.word.as_ref().len();
+        if self.prev < len || self.prev == 0 {
+            let w = self.word.split(self.prev..len, true);
+            // TODO: shouldn’t this be just len?
```
mgeisler, line 259: I could have added a comment to explain this: with just

```
self.prev = len;
```

you get an infinite loop when `len == 0` since `self.prev == 0` stays true after every call to `next`.

mgeisler: If you remove the `+1` on `master`, you'lll see that `cargo test` hangs.

examples/style.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn penalty_width(&self) -> usize {
+        if self.hyphen {
+            1
+        } else {
+            0
+        }
+    }
+}
+
+impl textwrap::word_splitters::Splittable for StyledWord<'_> {
```
mgeisler, line 68: It's a bit of a shame that the example doesn't show why there are two traits :-) I love the example in itself, it's super great at demonstrating the concept of wrapping not-just-plain-text. However, it would be nice if it would exploit the two traits better.

Will you be having different structs in `genpdf`, one for `Fragment` and another for `Splittable`? if not, then I would prefer to keep the number of concepts low and add a `split` method to `Fragment`.

mgeisler: Ah, sorry, you already explained that you have a `StyledWord` struct for the unmeasured case and a `StyledFragment` for the pre-split and measured words.

src/word_splitters.rs
robinkrahl: I see, thanks.  Will remove the TODO.

examples/style.rs
robinkrahl: Exactly, the distinction is especially relevant if the width computation is non-trivial, which is typically the case for scenarios other than the terminal.  I could add an example that produces a PDF file, but I think that would be too complex to be useful as an example for `textwrap`.  Maybe we can have an example that produces an SVG image?

mgeisler: You're right that a full-blown PDF seems unnecessary — could you instead pretend that you need two structs in the `style` example? I believe you're using `len()` on the strings, which is cheating ever so slightly :-)

It might look a bit arbitrary, but for educational purposes, I think we're allowed to exaggerate.

robinkrahl: Yeah, we can do that.

PR #403, opened 2021-07-01
Configure required features for examples

robinkrahl:
Instead of manually checking for required features in the hyphenation
and termwidth examples, we can configure the required features in
Cargo.toml.  This makes the examples easier to read and maintain.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,14 @@ license = "MIT"
 edition = "2018"
 exclude = [".github/", ".gitignore", "benches/", "examples/", "fuzz/", "images/"]
 
+[[example]]
+name = "hyphenation"
+required-features = ["hyphenation"]
+
+[[example]]
+name = "termwidth"
+required-features = ["terminal_size"]
+
 [package.metadata.docs.rs]
 all-features = true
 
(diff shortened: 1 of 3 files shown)


mgeisler:
Oh yes, that would be much better!

robinkrahl:
Hmm, it seems to break the CI.  I’ll try to fix it.

robinkrahl:
I’ve added some imports that I had accidentally removed, now it’s working again.

mgeisler:
Thanks, that's much nicer!

mgeisler:
Hi Robin,

Just a FYI in case you run into this too... :-) I got a weird error when trying to publish the crate after this:

```
% cargo publish --dry-run
    Updating crates.io index
   Packaging textwrap v0.15.0 (/home/mg/src/textwrap)
   Verifying textwrap v0.15.0 (/home/mg/src/textwrap)
error: failed to verify package tarball

Caused by:
  failed to parse manifest at `/home/mg/src/textwrap/target/package/textwrap-0.15.0/Cargo.toml`

Caused by:
  can't find `hyphenation` example at `examples/hyphenation.rs` or
  `examples/hyphenation/main.rs`. Please specify example.path if you
  want to use a non-default path.
```

I could fix it by specifying the path explicitly: #444.

PR #406, opened 2021-07-06
Rename `Fragment` to `MeasuredFragment`

mgeisler:
This name better matches the new `SplittableFragment` trait in #402.


mgeisler:
This is obsolete now, I'm pretty happy with the current `Fragment` trait.

PR #407, opened 2021-07-07
Add checks for README examples and fix typo

GuillaumeGomez:
You had a small typo in one of your examples, so I fixed it and added a check for it. I added the `doc-comment` dependency as a dev-dependency.

PR #408, opened 2021-07-08
Wrap README doctest in separate module

mgeisler:
We’ve gotten doctests for the README in #407. This PR changes the output on test failures from

    test src/lib.rs - (line 267) ... FAILED

to

    test src/lib.rs - readme_doctest (line 268) ... FAILED

which is slightly more informative. The line numbers are still off since I imagine they’re a combination of the line number in `lib.rs` and `README.md`. However, at least we now have an indication of where the error comes from. Finding the broken example should be easy then.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -185,7 +185,9 @@
 #![allow(clippy::redundant_field_names)]
 
 #[cfg(all(doctest, feature = "hyphenation"))]
-doc_comment::doctest!("../README.md");
+mod readme_doctest {
+    doc_comment::doctest!("../README.md");
+}
 
 use std::borrow::Cow;
 


mgeisler:
@GuillaumeGomez Thanks for #407 — I imagine others would also like to wrap the doctest in a separate module to get a better indication of where the error comes from. Do you think the `doctest!` macro could generate such a module automatically?

I think it could derive the module name from the path of the file to be tested — though I'm unsure how to implement this in a macro.

GuillaumeGomez:
It's already supported:

```rust
// If you want to have a name for your tests:
doc_comment::doctest!("../README.md", another);
```

I took this example from the docs: https://docs.rs/doc-comment/0.3.3/doc_comment/macro.doctest.html (yep, maybe not so easy to see though...).

mgeisler:
Oh, silly me! I only browsed the front page of the documentation very briefly on https://lib.rs/doc-comment, so I didn't get to the actual documentation of the `doctest!` macro. Fixed in #409.

PR #420, opened 2021-12-31
Prevent overflows by using `u32` instead of `usize` to represent the line width

mgeisler:
Before, we used `usize` for the line width, now we use an integer type with a fixed width of 32 bits.

Using an `usize` for the line width can make the optimal-fit wrapping algorithm overflow when it tries to compute the optimal wrapping cost. The problem is that the algorithm works with integer values formed by

    (line_width - target_width)**2

and when `line_width` is near `usize::max_value()`, this computation can overflow an `usize`. By limiting the line width to an `u32`, we can do the internal computations with `u128` and avoid the overflows.

The lack of overflows is checked by the new fuzz tests: `wrap_first_fit.rs` and `wrap_optimal_fit.rs`. These fuzz tests generate completely random segments with arbitrary widths and checks that they can be wrapped without overflow. The new fuzz tests find the original overflow bugs within a split second.

----

This is a continuation of #418, just with a better branch name.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,11 +134,17 @@ jobs:
       - name: Build fuzz targets
         run: cargo fuzz build
 
-      - name: Fuzz test
+      - name: Fuzz test fill_first_fit
         run: cargo fuzz run fill_first_fit -- -max_total_time=30
 
-      - name: Minimize fuzz corpus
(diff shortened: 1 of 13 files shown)


mgeisler:
Please also see #421 which solves this in a more flexible way.

mgeisler:
Superseded by #421.

Issue #453, opened 2022-06-02
No explicit control over CR LF line ending


edio:
[Helix editor](https://github.com/helix-editor) recently added a feature to reflow a piece of text; _helix_ uses _textwrap_ for this

There's an issue, where reflow does not work correctly if `CRLF` sequence (`\r\n`) is used as a line terminator
https://github.com/helix-editor/helix/issues/2645

_helix_ does not support `CR` ending (pre macosx), but I guess, the issue would manifest with it as well.

As far as I can see, _textwrap_ does not account for different line ending, there's no configuration option with which we could tune _textwrap_ behavior.

Did I miss anything? Is there a way to support `CRLF` in _textwrap_?

If not, I'd be willing to contribute a fix.
My initial idea is to introduce another option into the `Options` struct, named `line_ending` (or `line_terminator` as a fancier alternative). To keep things tight, I'd express it as `enum` (we can always create something like `Custom` should we want to support an arbitrary byte sequence later, but I highly doubt that). How does that sound?

mgeisler:
Hey @edio, thanks for opening the issue. Super exciting that textwrap is used in the editor!

You're not missing anything, textwrap is hard-coded to split on `\n` in [`textwrap::wrap`](https://github.com/mgeisler/textwrap/blob/6fa2cd848a6611e10f74ec9ff0a89c8b1bc7ce80/src/lib.rs#L918). I see you're using `textwrap::refill` and that is also hard-coded to use `\n`.

I've been thinking of `\n` as the _logical_ way to separate lines and `\r\n` as more of a fileformat _encoding_ of the logical `\n` characters. That is, my opinion is that programs should use `\n` throughout their processing and then encode the string to the target file format (e.g., do a `.replace("\n", "\r\n")` call at the point where the data leaves the program).

Searching a bit, it seems that this way of thinking comes from how languages like [C++ handle text mode files](https://stackoverflow.com/a/213977/110204). Rust doesn't [have a text mode](https://github.com/rust-lang/rust/issues/19403), so I think my way of thinking is outdated :-D

Now, I see a few different approaches for handling this:

* We could either push the problem back to the caller: demand that you use `\n` delimited string with Textwrap. That matches my way of thinking above. Easiest solution to implement for Textwrap :smile: 
* Alternatively, we could split on a custom line ending as you suggest.
[...]

PR #454, opened 2022-06-04
Add configurable LineEnding

edio:
This is a quick demonstration of how configurable line ending support would look like #453 .

Immediately I see an issue with `unfill` which now requires `LineEnding` parameter, degrading experience for the clients who do not wish to deal with configurable line endings.

Perhaps, we could have two fns: `unfill_with_endings(&str, LineEnding)` and `unfill(&str) { unfill_with_endings(str, LineEnding::LF) }`

Alternatives to consider
- using `lines()` in `unfill` (trimming trailing `\r\n` could be implemented as discarding trailing empty lines)
- autodetecting the end of the line by examining the input and matching against a set of known patterns
 
My gut feeling is that explicit control over the behavior is better (unicode defines multiple other options over the well-known `CR` and `LF`), but I do not think this is a strong argument considering `lines()` only recognizes `\n` and `\r\n`.

If the overall approach (or any of the alternatives) seems reasonable, I'll work towards finalizing this by providing tests, documentation, etc.. Let me know please.

### DIFF
--- a/benches/unfill.rs
+++ b/benches/unfill.rs
@@ -1,0 +1,23 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+
+pub fn benchmark(c: &mut Criterion) {
+    let words_per_line = [
+        5, 10, 15, 5, 5, 10, 5, 5, 5, 10, // 10 lines
+        10, 10, 5, 5, 5, 5, 15, 10, 5, 5, // 20 lines
+        10, 5, 5, 5, 15, 10, 10, 5, 5, 5, // 30 lines
+        15, 5, 5, 10, 5, 5, 5, 15, 5, 10, // 40 lines
+        5, 15, 5, 5, 15, 5, 10, 10, 5, 5, // 50 lines
+    ];
+    let mut text = String::new();
+    for (line_no, word_count) in words_per_line.iter().enumerate() {
+        text.push_str(&lipsum::lipsum_words_from_seed(*word_count, line_no as u64));
+        text.push('\n');
+    }
+    text.push_str("\n\n\n\n");
+    assert_eq!(text.len(), 2650); // The size for reference.
+
+    c.bench_function("unfill", |b| b.iter(|| textwrap::unfill(&text)));
+}
+
+criterion_group!(benches, benchmark);
+criterion_main!(benches);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -224,7 +224,7 @@ pub mod core;
 pub struct Options<'a> {
     /// The width in columns at which the text will be wrapped.
     pub width: usize,
-    /// TODO doc
+    /// Line ending used for breaking lines.
     pub line_ending: LineEnding,
     /// Indentation used for the first line of output. See the
     /// [`Options::initial_indent`] method.
@@ -341,7 +341,21 @@ impl<'a> Options<'a> {
         Self::new(termwidth())
     }
 
-    /// TODO
+    /// Change [`self.line_ending`]. This specifies which of the
+    /// supported line endings should be used to break the lines of the
+    /// input text.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use textwrap::{refill, LineEnding, Options};
+    ///
+    /// let options = Options::new(15).line_ending(LineEnding::CRLF);
+    /// assert_eq!(refill("This is a little example.", options),
+    ///            "This is a\r\nlittle example.");
+    /// ```
+    ///
+    /// [`self.line_ending`]: #structfield.line_ending
     pub fn line_ending(self, line_ending: LineEnding) -> Self {
         Options {
             line_ending,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -601,10 +601,16 @@ where
 /// textwrap: a small library for wrapping text.
 /// ```
 ///
-/// In addition, it will recognize a common prefix among the lines.
+/// In addition, it will recognize a common prefix and a common line
+/// ending among the lines.
+///
 /// The prefix of the first line is returned in
 /// [`Options::initial_indent`] and the prefix (if any) of the the
 /// other lines is returned in [`Options::subsequent_indent`].
+///
+/// Line ending is returned in [`Options::line_ending`]. If line ending
+/// can not be confidently detected (mixed or no line endings in the
+/// input), [`LineEnding::LF`] will be returned.
 ///
 /// In addition to `' '`, the prefixes can consist of characters used
 /// for unordered lists (`'-'`, `'+'`, and `'*'`) and block quotes
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -648,6 +648,7 @@ where
 /// assert_eq!(text, "This is an example of a list item.\n");
 /// assert_eq!(options.initial_indent, "* ");
 /// assert_eq!(options.subsequent_indent, "  ");
+/// assert_eq!(options.line_ending, LineEnding::LF);
 /// ```
 pub fn unfill(text: &str) -> (String, Options<'_>) {
     let line_ending_pat: &[_] = &['\r', '\n'];
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -214,7 +214,7 @@ pub use word_splitters::WordSplitter;
 pub mod wrap_algorithms;
 pub use wrap_algorithms::WrapAlgorithm;
 
-pub mod line_ending;
+mod line_ending;
 pub use line_ending::LineEnding;
 
 pub mod core;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -650,7 +650,8 @@ where
 /// assert_eq!(options.subsequent_indent, "  ");
 /// ```
 pub fn unfill(text: &str) -> (String, Options<'_>) {
-    let trimmed = text.trim_end_matches(&['\r', '\n']);
+    let line_ending_pat: &[_] = &['\r', '\n'];
+    let trimmed = text.trim_end_matches(line_ending_pat);
     let prefix_chars: &[_] = &[' ', '-', '+', '*', '>', '#', '/'];
 
     let mut options = Options::new(0);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -656,7 +656,7 @@ pub fn unfill(text: &str) -> (String, Options<'_>) {
     let prefix_chars: &[_] = &[' ', '-', '+', '*', '>', '#', '/'];
 
     let mut options = Options::new(0);
-    for (idx, line) in trimmed.split('\n').enumerate() {
+    for (idx, line) in trimmed.lines().enumerate() {
         options.width = std::cmp::max(options.width, core::display_width(line));
         let without_prefix = line.trim_start_matches(prefix_chars);
         let prefix = &line[..line.len() - without_prefix.len()];
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -922,7 +922,7 @@ where
 {
     let options: Options = width_or_options.into();
 
-    let line_ending_pat = options.line_ending.as_str();
+    let line_ending_str = options.line_ending.as_str();
 
     let initial_width = options
         .width
@@ -932,7 +932,7 @@ where
         .saturating_sub(core::display_width(options.subsequent_indent));
 
     let mut lines = Vec::new();
-    for line in text.split(line_ending_pat) {
+    for line in text.split(line_ending_str) {
         let words = options.word_separator.find_words(line);
         let split_words = word_splitters::split_words(words, &options.word_splitter);
         let broken_words = if options.break_words {
@@ -1831,6 +1831,14 @@ mod tests {
     #[test]
     fn unfill_whitespace() {
         assert_eq!(unfill("foo   bar").0, "foo   bar");
+    }
+
+    #[test]
+    fn refill_convert_crlf() {
+        assert_eq!(
+            refill("foo\nbar\n", Options::new(5).line_ending(LineEnding::CRLF)),
+            "foo\r\nbar\n",
+        );
     }
 
     #[test]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -214,6 +214,9 @@ pub use word_splitters::WordSplitter;
 pub mod wrap_algorithms;
 pub use wrap_algorithms::WrapAlgorithm;
 
+pub mod line_ending;
+pub use line_ending::LineEnding;
+
 pub mod core;
 
 /// Holds configuration options for wrapping and filling text.
@@ -221,6 +224,8 @@ pub mod core;
 pub struct Options<'a> {
     /// The width in columns at which the text will be wrapped.
     pub width: usize,
+    /// TODO doc
+    pub line_ending: LineEnding,
     /// Indentation used for the first line of output. See the
     /// [`Options::initial_indent`] method.
     pub initial_indent: &'a str,
@@ -248,6 +253,7 @@ impl<'a> From<&'a Options<'a>> for Options<'a> {
     fn from(options: &'a Options<'a>) -> Self {
         Self {
             width: options.width,
+            line_ending: options.line_ending,
             initial_indent: options.initial_indent,
             subsequent_indent: options.subsequent_indent,
             break_words: options.break_words,
@@ -268,12 +274,13 @@ impl<'a> Options<'a> {
     /// Creates a new [`Options`] with the specified width. Equivalent to
     ///
     /// ```
-    /// # use textwrap::{Options, WordSplitter, WordSeparator, WrapAlgorithm};
+    /// # use textwrap::{LineEnding, Options, WordSplitter, WordSeparator, WrapAlgorithm};
     /// # let width = 80;
     /// # let actual = Options::new(width);
     /// # let expected =
     /// Options {
     ///     width: width,
+    ///     line_ending: LineEnding::LF,
     ///     initial_indent: "",
     ///     subsequent_indent: "",
     ///     break_words: true,
@@ -289,6 +296,7 @@ impl<'a> Options<'a> {
     /// }
     /// # ;
     /// # assert_eq!(actual.width, expected.width);
+    /// # assert_eq!(actual.line_ending, expected.line_ending);
     /// # assert_eq!(actual.initial_indent, expected.initial_indent);
     /// # assert_eq!(actual.subsequent_indent, expected.subsequent_indent);
     /// # assert_eq!(actual.break_words, expected.break_words);
@@ -301,6 +309,7 @@ impl<'a> Options<'a> {
     pub const fn new(width: usize) -> Self {
         Options {
             width,
+            line_ending: LineEnding::LF,
             initial_indent: "",
             subsequent_indent: "",
             break_words: true,
@@ -330,6 +339,14 @@ impl<'a> Options<'a> {
     #[cfg(feature = "terminal_size")]
     pub fn with_termwidth() -> Self {
         Self::new(termwidth())
+    }
+
+    /// TODO
+    pub fn line_ending(self, line_ending: LineEnding) -> Self {
+        Options {
+            line_ending,
+            ..self
+        }
     }
 
     /// Change [`self.initial_indent`]. The initial indentation is
@@ -428,6 +445,7 @@ impl<'a> Options<'a> {
     pub fn word_separator(self, word_separator: WordSeparator) -> Options<'a> {
         Options {
             width: self.width,
+            line_ending: self.line_ending,
             initial_indent: self.initial_indent,
             subsequent_indent: self.subsequent_indent,
             break_words: self.break_words,
@@ -446,6 +464,7 @@ impl<'a> Options<'a> {
     pub fn wrap_algorithm(self, wrap_algorithm: WrapAlgorithm) -> Options<'a> {
         Options {
             width: self.width,
+            line_ending: self.line_ending,
             initial_indent: self.initial_indent,
             subsequent_indent: self.subsequent_indent,
             break_words: self.break_words,
@@ -473,6 +492,7 @@ impl<'a> Options<'a> {
     pub fn word_splitter(self, word_splitter: WordSplitter) -> Options<'a> {
         Options {
             width: self.width,
+            line_ending: self.line_ending,
             initial_indent: self.initial_indent,
             subsequent_indent: self.subsequent_indent,
             break_words: self.break_words,
@@ -546,13 +566,16 @@ pub fn fill<'a, Opt>(text: &str, width_or_options: Opt) -> String
 where
     Opt: Into<Options<'a>>,
 {
+    let options = width_or_options.into();
+    let line_ending_str = options.line_ending.as_str();
+
     // This will avoid reallocation in simple cases (no
     // indentation, no hyphenation).
     let mut result = String::with_capacity(text.len());
 
-    for (i, line) in wrap(text, width_or_options).iter().enumerate() {
+    for (i, line) in wrap(text, options).iter().enumerate() {
         if i > 0 {
-            result.push('\n');
+            result.push_str(line_ending_str);
         }
         result.push_str(line);
     }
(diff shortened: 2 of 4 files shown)


edio:
@mgeisler , please let me know what do you think about the overall approach.
(and just in case, I'm totally ok if after looking at the code you decide that this approach does not fit)
Thanks for your time.

src/lib.rs
```diff
@@ -610,24 +633,25 @@ where
...
-    let trimmed = text.trim_end_matches('\n');
+pub fn unfill(text: &str, line_ending: LineEnding) -> (String, Options<'_>) {
+    let line_ending_pat = line_ending.as_str();
+    let trimmed = text.trim_end_matches(line_ending_pat);
     let prefix_chars: &[_] = &[' ', '-', '+', '*', '>', '#', '/'];
 
-    let mut options = Options::new(0);
-    for (idx, line) in trimmed.split('\n').enumerate() {
+    let mut options = Options::new(0).line_ending(line_ending);
+    for (idx, line) in trimmed.split(line_ending_pat).enumerate() {
```
mgeisler, line 654: What about auto-detecting the line ending here? Basically, if we find a `"\r\n"` in `text` then we use that line ending, otherwise we use `"\n"`.

I think it would be okay to say that `unfill` doesn't handle input with a mix of `\r\n` and `\n` line endings (if the caller is concerned about such cases, I would suggest the caller to normalize the input first).

src/line_ending.rs
```diff
@@ -0,0 +1,50 @@
...
+            _ => 1,
+        }
+    }
+
+    /// TODO
+    #[inline]
+    pub const fn as_str(&self) -> &'static str {
+        match self {
+            Self::CRLF => "\u{000D}\u{000A}",
+            Self::LF => "\u{000A}",
```
mgeisler, line 31: I'm not very good at remembering the hexadecimal ASCII values :-) Could we use `\n` instead here and above?

src/line_ending.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// TODO
+    CRLF,
+    /// TODO
+    LF,
+}
+
+impl LineEnding {
+    /// TODO
+    #[inline]
+    pub const fn len_chars(&self) -> usize {
```
mgeisler, line 19: This seems unused?

src/lib.rs
```diff
@@ -1768,7 +1795,15 @@ mod tests {
 
     #[test]
     fn unfill_whitespace() {
-        assert_eq!(unfill("foo   bar").0, "foo   bar");
+        assert_eq!(unfill("foo   bar", LineEnding::LF).0, "foo   bar");
+    }
+
+    #[test]
+    fn unfill_crlf() {
```
mgeisler, line 1802: If we do the auto-detection like I suggest, could you then also please add another test which uses `refill` to change the line endings from `\n` to `\r\n`?

That is, I believe this should work:

```rust
assert_eq(refill("foo\nbar\n", Options::new(5).line_ending(LineEndings::CRLF)),
          "foo\r\nbar\r\n");
```

This would show that we correctly use the line ending from the input option to `refill` to join the lines after we use `unfill` to get the text unwrapped.

src/lib.rs
```diff
@@ -214,6 +214,9 @@ pub use word_splitters::WordSplitter;
 pub mod wrap_algorithms;
 pub use wrap_algorithms::WrapAlgorithm;
 
+pub mod line_ending;
```
mgeisler, line 217: I think the module is small enough that we can just inline it into `lib.rs`. Or perhaps better: leave it as an internal module and then `pub use` the enum like you already do. Basically, since I don't think the module will contain anything which isn't also exported at the root level, I think we should not expose the module.

mgeisler:
Thanks so much for the PR, it looks really good! I've left a few comments, but overall I think this should work just fine.

mgeisler:
> Perhaps, we could have two fns: `unfill_with_endings(&str, LineEnding)` and `unfill(&str) { unfill_with_endings(str, LineEnding::LF) }`
> 
> Alternatives to consider
> 
> * using `lines()` in `unfill` (trimming trailing `\r\n` could be implemented as discarding trailing empty lines)
> * autodetecting the end of the line by examining the input and matching against a set of known patterns

Yeah, thanks for listing this out. My preference would be for auto-detecting the input line ending in `unfill`: the job of the function is to make sense of a piece of text without anyone telling it how wide the lines are, what the prefix is etc. So detecting the line ending used would consistent with that.

Of course, it could happen that there are no newlines in the input text: if so, then I think it's acceptable that the caller gets an `Options` back with `LineEnding::LF`. Basically, such an input string is ambiguous and Textwrap can default to `LineEnding::LF`.

edio:
@mgeisler , got you, thanks!

Not arguing, just want to explain my thinking: Rust's `lines()` is smart and efficient about the CRLF support.
It always splits by `\n` and immediately after the split it checks whether the last byte of the line is `\r`, and removes it, if that's the case. However `\r\n in text` would require full traversal over the input (and, btw, without a guarantee that it'll actually find it, so it is indeed full scal), and then another full traversal to actually do splits.

(also, ACK to all your comments, will get to the fixes ASAP, but next weekend may be busy for me, so I may get to it only next week. Sorry about the delay :pray: )

src/lib.rs
edio: I gave it a try and after some tinkering I made it work with mixed input and also made it 20% faster.

```
original ❯ cargo criterion --bench "unfill"
   Compiling textwrap v0.15.0 (/home/.../develop/textwrap)
    Finished bench [optimized] target(s) in 4.54s
  Executable benches/unfill.rs (target/release/deps/unfill-d9aa0a94ff4a34d4)
unfill                  time:   [19.092 us 19.105 us 19.117 us]                    
                        change: [+25.614% +25.731% +25.857%] (p = 0.00 < 0.05)
                        Performance has regressed.
updated ❯ cargo criterion --bench "unfill"
   Compiling textwrap v0.15.0 (/home/koiuo/develop/textwrap)
    Finished bench [optimized] target(s) in 4.57s
  Executable benches/unfill.rs (target/release/deps/unfill-d9aa0a94ff4a34d4)
unfill                  time:   [15.246 us 15.255 us 15.266 us]                    
                        change: [-19.845% -19.726% -19.585%] (p = 0.00 < 0.05)
                        Performance has improved.
```

I haven't updated a single existing test (except for added assertions) and haven't touched the prefix detection code. But I feel like we could make `unfill` work with multiple input paragraphs without significant code complication or performance degradation. If there's an interest I could give it a try after we are done with this PR.

src/lib.rs
edio: My recent change adds custom `NonEmptyLines` iterator to that module, which returns the line together with the detected line ending.

Do you still want this code to reside in lib.rs or should we keep it in a dedicated file?

src/lib.rs
```diff
@@ -634,16 +662,24 @@ pub fn unfill(text: &str) -> (String, Options<'_>) {
...
+    for (line, ending) in line_ending::NonEmptyLines(text) {
+        if unfilled.is_empty() {
             unfilled.push_str(&line[options.initial_indent.len()..]);
         } else {
             unfilled.push(' ');
             unfilled.push_str(&line[options.subsequent_indent.len()..]);
         }
+        match (detected_line_ending, ending) {
+            (None, Some(_)) => detected_line_ending = ending,
+            (Some(LineEnding::CRLF), Some(LineEnding::LF)) => detected_line_ending = ending,
```
edio, line 676: In case we have mixed line terminators I decided to give `LF` a higher priority. So once we've encountered `LF`, we return it in `Options`

My motivation is that `LF` is more "universal" than `CRLF`: splitting a mixed-terminator file with `CRLF` will fail to split some lines, while splitting with `LF` will not miss lines but will leave some garbage `CR`s. That said, `unfill` still handles both `LF` and `CRLF` with no garbage `CR`'s in case of mixed input.

src/lib.rs
mgeisler: Good point, I think you're right that we should keep it as a dedicated file. I don't think there is anything in the module which should be exported, so we can remove the `pub` here.

benches/unfill.rs
```diff
@@ -0,0 +1,23 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+
+pub fn benchmark(c: &mut Criterion) {
```
mgeisler, line 3: Thanks for writing this! :-)

src/lib.rs
```diff
@@ -607,7 +630,7 @@ where
 /// assert_eq!(options.subsequent_indent, "  ");
 /// ```
 pub fn unfill(text: &str) -> (String, Options<'_>) {
-    let trimmed = text.trim_end_matches('\n');
+    let trimmed = text.trim_end_matches(&['\r', '\n']);
     let prefix_chars: &[_] = &[' ', '-', '+', '*', '>', '#', '/'];
 
     let mut options = Options::new(0);
```
mgeisler, line 636: This comment should really be on the next line:

```rust
    for (idx, line) in trimmed.split('\n').enumerate() {
```

Just to double-check: splitting on `\n` still works because this also catches `\r\n`? You could consider moving the line ending detection logic up here and then use the detected line ending below. I think that should do the same, but it might be ever so slightly easier to understand.

src/lib.rs
```diff
@@ -891,6 +922,8 @@ where
 {
     let options: Options = width_or_options.into();
 
+    let line_ending_pat = options.line_ending.as_str();
```
mgeisler, line 925: I would call this `line_ending_str` like you did elsewhere (it's not really a pattern).

src/line_ending.rs
```diff
@@ -0,0 +1,67 @@
...
+            _ => Result::Err(()),
+        }
+    }
+}
+
+/// TODO
+#[derive(Debug, Clone, Copy)]
+pub struct NonEmptyLines<'a>(pub &'a str);
+
+impl<'a> Iterator for NonEmptyLines<'a> {
```
mgeisler, line 43: This is really neat!

src/line_ending.rs
```diff
@@ -0,0 +1,67 @@
...
+            "\u{000D}\u{000A}" => Result::Ok(LineEnding::CRLF),
+            "\u{000A}" => Result::Ok(LineEnding::LF),
+            _ => Result::Err(()),
+        }
+    }
+}
+
+/// TODO
+#[derive(Debug, Clone, Copy)]
+pub struct NonEmptyLines<'a>(pub &'a str);
```
mgeisler, line 41: This can probably be `pub(crate)` since it's only used internally in `unfill`?

mgeisler:
> Not arguing, just want to explain my thinking: 

Thanks for explaining!

> Rust's `lines()` is smart and efficient about the CRLF support. It always splits by `\n` and immediately after the split it checks whether the last byte of the line is `\r`, and removes it, if that's the case. However `\r\n in text` would require full traversal over the input (and, btw, without a guarantee that it'll actually find it, so it is indeed a full scan), and then another full traversal to actually do splits.

Right, that's very clever! I see you implemented equal cleverness now in a very elegant iterator, which is awesome!

> (also, ACK to all your comments, will get to the fixes ASAP, but next weekend may be busy for me, so I may get to it only next week. Sorry about the delay :pray: )

This is open source, there are no deadlines! We all work on this when we have time and energy :smiley: Thank you for the great PR!

src/lib.rs
edio: `LineEnding` enum should be exported I think (we allow clients to set it in `Options`)

src/lib.rs
edio: Will do. As I said, I might not be able to finish it this week, but ultimately I intend to ship this with tests and documentation 

- [ ] mention `LineEnding::LF` is default in case of mixed input in the docs

src/lib.rs
edio: > That is, I believe this should work

Except that we do not modify trailing line endings, hence it'll be 
```
"foo\r\nbar\n"
```
On one hand it makes sense (the lib's focus is text wrapping, not CRLF conversion), on the other it is a bit unexpected.

mgeisler: Hmm... can you remind me why I decided to not modify the trailing line endings? :smile: 

Perhaps it's because I made `unfill` so simple: it expects a single wrapped paragraph and so a string with `\n\n` doesn't really fit into that model. But now that I think (briefly) about it again, it seems to me that it would be equally consistent to collapse the line endings (if they exist). So `"foo\n\n\n"` becomes `"foo\n"` and `"foo"` stays `"foo"`. But I might be missing something?

src/lib.rs
mgeisler: Yep, it should be exported, but this can be done via the `pub use` in `lib.rs`. If you do that and you don't expose the `line_ending` module itself, the enum behaves exactly as if it had been defined in `lib.rs`. You can see that with [`WordSeparator`](https://docs.rs/textwrap/latest/textwrap/enum.WordSeparator.html) which shows up in the crate root, but is actually defined in a private module.

src/lib.rs
edio: That's a good idea.

When working on this PR I wanted to minimize changes to the existing codebase, hence I wanted to leave the common prefix detection code intact (since my feature should not directly impact it). Hence the result. I tried to depict the data flow on the graph below

```mermaid
graph LR;
    input --> A & C
    subgraph loop [same loop]
      direction TB;
      C & B;
    end
    A[fold_prefix] --> B[fold_lines];
    C[fold_ending] --> D[Options];
    B --> E[&str];
    subgraph result [result]
      D & E
    end
```

We could of course do it like you ask

```mermaid
graph LR;
    input --> A & C
    subgraph loop [same loop]
      direction TB;
      A & C;
    end
    A[fold_prefix] --> B[fold_lines];
    C[fold_ending] --> D[Options];
    B --> E[&str];
    subgraph result [result]
      D & E
    end
```

We'll need new iterator to perform two fold in one pass, it'll contain all the logic (still the same complexity just moved around).
We'll still need two full passes over the input unfortunately, because we need the result of the `fold_prefix` before we do `fold_lines`. 

I'm sharing all this to prove that I've done my homework and thought about this :-) and with that in mind I'd prefer to leave the code as it is in this PR. I'd prefer this new feature to be as narrowly scoped as possible, and I'd be happy to start re-factoring in a new PR.

src/lib.rs
edio: As a user of `unfill` I'd prefer it to do all the magic and do just what I want.

From that point of view collapsing new lines seems preferrable to current behavior.

If we implement multiple paragraphs handling (as I mentioned somewhere above), then this behavior would seem the most consistent to me.

1. `unfill` would process a paragraph as it is doing now
2. as soon as it detects an empty line, it would reset all the internal state and repeat the processing on the remainder of the input

Any empty lines (including trailing line breaks), would then be collapsed and replaced by a common auto-detected value.

edio: Btw, about `unfill` doing all the magic: I think detecting more prefixes than just common list markers would be very useful. Particularly, it'd be great if unfill understood that it operates on a comment. Just stumbled on this while editing this PR in _helix_ editor (tell me about eating my own dogfood :smile:  )

src/lib.rs
```diff
@@ -214,13 +214,18 @@ pub use word_splitters::WordSplitter;
...
+pub use line_ending::LineEnding;
+
 pub mod core;
 
 /// Holds configuration options for wrapping and filling text.
 #[derive(Debug, Clone)]
 pub struct Options<'a> {
     /// The width in columns at which the text will be wrapped.
     pub width: usize,
+    /// TODO doc
```
mgeisler, line 227: (Still need a docstring here.)

src/lib.rs
```diff
@@ -332,6 +341,14 @@ impl<'a> Options<'a> {
         Self::new(termwidth())
     }
 
+    /// TODO
```
mgeisler, line 344: Docstring needed.

src/line_ending.rs
```diff
@@ -0,0 +1,117 @@
+//! Line ending detection and conversion.
+
+use std::fmt::{Debug, Formatter};
+use std::str::FromStr;
+
+/// Supported line endings. Like in the Rust's standard library, two
```
mgeisler, line 6: This should be

> Like in the Rust standard library

or

> Line in Rust's standard library

I prefer the first form without the apostrophe since it seems simpler to a non-native speaker like me :-)

src/line_ending.rs
```diff
@@ -0,0 +1,117 @@
...
+/// line endings are supported: `\r\n` and `\n`
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum LineEnding {
+    /// _Carriage return and line feed_ – a line ending sequence
+    /// historically used in _Windows_. Corresponds to the sequence
```
mgeisler, line 11: I don't think we have a convention of putting product names in italic in the documentation, so please change `_Windows_` to `Windows` (and below for Unix).

src/line_ending.rs
```diff
@@ -0,0 +1,117 @@
...
+    #[inline]
+    pub const fn as_str(&self) -> &'static str {
+        match self {
+            Self::CRLF => "\r\n",
+            Self::LF => "\n",
+        }
+    }
+}
+
+impl FromStr for LineEnding {
```
mgeisler, line 43: This is not used and I don't think we need it. Please drop it until someone has a use case for it :-)

src/line_ending.rs
```diff
@@ -0,0 +1,117 @@
...
+            return result;
+        } else {
+            return None;
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::line_ending::NonEmptyLines;
```
mgeisler, line 92: Just as a FYI, I normally use `use super::*` in the test modules to conveniently pull in everything from the parent scope. This is fine too in case you like the imports to be more precise.

src/lib.rs
mgeisler: That sounds good, I like it a lot that you try to keep the scope of the new feature narrow! It's hard to review PRs if they move random lines around and refactor the code while also making the actual changes :-)

src/lib.rs
mgeisler: > As a user of `unfill` I'd prefer it to do all the magic and do just what is the most reasonable (if I need any advanced tailored behavior I'd probably implement it myself).
> 
> From that point of view collapsing new lines seems preferrable to current behavior.

I think I agree: you're saying that `unfill` should not keep training newlines like it does now. If so, then I agree :-)

> 1. `unfill` would process a paragraph as it is doing now
> 2. as soon as it detects an empty line, it would reset all the internal state and repeat the processing on the remainder of the input
> 
> Any empty lines (including trailing line breaks), would then be collapsed and replaced by a common auto-detected value.

I guess the multi-paragraph logic of `unfill` would be:

* a single `\n` is seen as a continuation of the current paragraph. It is removed and the text from both lines go into the current line.
* two or more `\n` is seen as a paragraph separator. The newlines are added to the unfolded text followed by the current line.

So

```
this is
the first
paragraph


second paragraph
still
second

third paragraph
```

becomes 

```
this is the first paragraph


second paragraph still second

third paragraph
```

Here `\n` and `\r\n` should be treated the same: if the input has a trailing `\r\n` or a `\r\n\r\n` paragraph separator, then the unfolded string should use `\r\n` as well. It will also indicate that this is the line ending by putting it `LineEnding::CRLF` into the `line_ending` field.

The caller (`refill`) can then split on `options.line_ending.as_str()` to get the individual _paragraphs_. It can fold them using the _new_ line ending and thus we should end up with a line ending conversion.

Does that seem reasonable?

src/lib.rs
edio: ACK, sorry, missed it entirely

mgeisler:
Hey @edio, thanks for all the great updates on the PR.

The build error on Rust 1.56 looks the same as what I had in another project: https://github.com/mgeisler/lipsum/pull/84. I think you just need to be explicit with the type of the slice (for some reason which I don't understand).

src/lib.rs
mgeisler: All good, I just left comments so that we won't forget :-)

edio:
@mgeisler , thank you for the fix for the Rust 1.56 compatibility issue. It'd took me ages to figure that out.

I believe I adressed all your requests for changes, but please do not hesitate to point to other problems if you can think of any.

And thanks again for your time and help on this PR.

src/lib.rs
edio: My thinking is the same, I believe we are almost¹ on the same page regarding this. Thanks for putting it into words.

> The caller (`refill`) can then split on options.line_ending.as_str() to get the individual paragraphs.

¹ - If you meant to say `unfill` here, then I agree with you. Otherwise I think I may be missing something.

src/lib.rs
```diff
@@ -594,7 +637,7 @@ where
 /// # Examples
 ///
 /// ```
-/// use textwrap::unfill;
+/// use textwrap::{LineEnding, unfill};
```
mgeisler, line 640: I believe you wanted to add `assert_eq!(options.line_ending, LineEnding::CR)` below so that this import is used.

mgeisler (approved):
Thanks for all the updates!

mgeisler:
> thank you for the fix for the Rust 1.56 compatibility issue. It'd took me ages to figure that out.

Yeah, that was a weird case for me too :smile:

src/line_ending.rs
```diff
@@ -0,0 +1,102 @@
...
+    CRLF,
+    /// _Line feed_ – a line ending historically used in Unix.
+    ///  Corresponds to the ASCII control character `0x0A` or `\n`
+    LF,
+}
+
+/// Returned when attempted creating [`LineEnding`] value from an
+/// unsupported `&str` value
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct UnsupportedLineEnding;
```
mgeisler, line 21: Could you remove this struct now that it's unused?

mgeisler (approved):

src/lib.rs
```diff
@@ -1697,6 +1752,34 @@ mod tests {
...
+    #[test]
+    fn unfill_no_new_line() {
+        let (text, options) = unfill("foo bar");
+        assert_eq!(text, "foo bar");
+        assert_eq!(options.width, 7);
+        assert_eq!(options.line_ending, LineEnding::LF);
+    }
+
+    #[test]
+    fn unfill_simple_crlf() {
```
edio, line 1767: All tests where CRLF is invloved fail during the CI. Obviously, I don't see any issues locally.

From the assertion error it seems that `\r` is not removed. I will try to reproduce this behavior locally using the same container image as we use in the CI pipeline.

mgeisler: > All tests where CRLF is invloved fail during the CI. Obviously, I don't see any issues locally.

Yeah, I could also not reproduce it at first until I tried with `cargo test --no-default-features`.

The problem is differing behavior of `display_width` depending on the enabled features. If the `unicode-width` feature is disabled, `display_width` will naively count each `char` as being 1 column wide. So we get

```
% cargo test --no-default-features tests::unfill_simple_crlf
  ...
test tests::unfill_simple_crlf ... FAILED

failures:

---- tests::unfill_simple_crlf stdout ----
[src/core.rs:185] &text = "foo\r"
[src/core.rs:185] width = 4
[src/core.rs:185] &text = "bar"
[src/core.rs:185] width = 3
thread 'tests::unfill_simple_crlf' panicked at 'assertion failed: `(left == right)`
  left: `4`,
 right: `3`', src/lib.rs:1771:9
```

The input string is `"foo\r\nbar"` and because we split on `\n` when counting the line width, we end up with the `"foo\r"` line.

edio: Thanks for your help.

As a simple and straightforward fix I decided to try `lines()` instead of `split('\n')` and to my surprise (I assumed `split` is there for a reason) I got better performance in the `unfill` benchmark (~19% difference in the same fashion I've got after implementing `NonEmptyLines` iterator). I fixed my cpu frequency to avoid noise and checked multiple times. The results are consistent.

mgeisler: Great! I think I used `\n` because I was afraid to eat `\r` characters by accident. But this is now handled with the auto-detection.

mgeisler:
Thanks for all the hard work on this!

PR #461, opened 2022-07-06
Update to terminal_size 0.2.

sunfishcode:
This updates the terminal_size optional dependency to the latest
version, 0.2.

I've tested that the termwidth example works.


mgeisler:
Thank you!

PR #467, opened 2022-09-13
Fix `unfill` and `refill` crashes

### LINKED ISSUE #466: Bugs with refill
1. Running refill with certain inputs will crash.
2. Running refill, once, then again on the results, with the same options, doesn't always produce the same results.

Wrote some fuzz targets to test:
1. 
```rust
#![no_main]
use libfuzzer_sys::fuzz_target;

fuzz_target!(|input: (String, usize)| {
    let _ = textwrap::refill(&input.0, input.1);
});

```
2.
```rust
#![no_main]
use libfuzzer_sys::fuzz_target;
use textwrap::refill;

fuzz_target!(|input: String| {
    let len = 20;
    let refilled = refill(&input, len);
    let refilled_again = refill(&refilled, len);
    assert_eq!(refilled, refilled_again);
});
```

Downstream issue in Helix: helix-editor/helix#3734

mgeisler:
This fixes two bugs in `unfill`, both to do with the way we were slicing strings in cases where the input ended with a combination of `\r` and `\n`.

Fixes #466.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,6 +120,8 @@ jobs:
           - wrap_first_fit
           - wrap_optimal_fit
           - wrap_optimal_fit_usize
+          - unfill
+          - refill
 
     steps:
       - name: Checkout repository
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -47,3 +47,15 @@ name = "wrap_optimal_fit_usize"
 path = "fuzz_targets/wrap_optimal_fit_usize.rs"
 test = false
 doc = false
+
+[[bin]]
+name = "refill"
+path = "fuzz_targets/refill.rs"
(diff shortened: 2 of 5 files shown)


mgeisler:
Hi @edio, do you remember why the iterator from #454 only iterated over non-empty lines?

edio:
@mgeisler , sorry for the late response.

It's hard to recall now, but I guess it was to allow _unfill_ operate on multiple paragraphs

```
paragraph1
text text

pargraph2
text text
```
would be unfilled into
```
paragraph1 text text pargraph2 text text
```

That's the only explanation I can think of right now :see_no_evil:

mgeisler:
Hi @edio,

Thanks, multiple paragraphs is something we might want to support one day... I think that would name `unfill` consistent with how `fill` works.

PR #490, opened 2022-11-02
Add `tab_width` support

mtoohey31:
This merge request is an updated version of #429. It allows the width of a tab to be configured for a variety of this crate's functions, instead of the width being fixed at zero, as it currently is.

### DIFF
--- a/src/columns.rs
+++ b/src/columns.rs
@@ -1,7 +1,6 @@
 //! Functionality for wrapping text into columns.
 
-use crate::core::display_width;
-use crate::{wrap, Options};
+use crate::{core, wrap, Options};
 
 /// Wrap text into columns with a given total width.
 ///
@@ -23,9 +22,9 @@ use crate::{wrap, Options};
 /// # let columns = 2;
 /// # let options = textwrap::Options::new(80);
 /// let inner_width = options.width
-///     - textwrap::core::display_width(left_gap)
-///     - textwrap::core::display_width(right_gap)
-///     - textwrap::core::display_width(middle_gap) * (columns - 1);
+///     - textwrap::core::display_width(left_gap, options.tab_width)
+///     - textwrap::core::display_width(right_gap, options.tab_width)
+///     - textwrap::core::display_width(middle_gap, options.tab_width) * (columns - 1);
 /// let column_width = inner_width / columns;
 /// ```
 ///
@@ -74,6 +73,8 @@ where
     assert!(columns > 0);
 
     let mut options: Options = total_width_or_options.into();
+    let tab_width = options.tab_width;
+    let display_width = |text| core::display_width(text, tab_width);
 
     let inner_width = options
         .width
@@ -189,5 +190,22 @@ mod tests {
     #[should_panic]
     fn wrap_columns_panic_with_zero_columns() {
         wrap_columns("", 0, 10, "", "", "");
+    }
+
+    #[test]
+    fn wrap_columns_with_tabs() {
+        let options = Options::new(23).tab_width(4);
+
+        #[cfg(feature = "smawk")]
+        let expected = vec!["|hello     |is\tlong|", "|this      |yeah      |"];
+        #[cfg(all(not(feature = "smawk"), feature = "unicode-linebreak"))]
+        let expected = vec!["|hello     |long      |", "|this\tis|yeah      |"];
+        #[cfg(not(any(feature = "smawk", feature = "unicode-linebreak")))]
+        let expected = vec!["|hello\tt|\tlong  |", "|his\tis |\tyeah  |"];
+
+        assert_eq!(
+            wrap_columns("hello\tthis\tis\tlong\tyeah", 2, options, "|", "|", "|"),
+            expected
+        )
     }
 }
--- a/src/core.rs
+++ b/src/core.rs
@@ -61,8 +61,12 @@ pub(crate) fn skip_ansi_escape_sequence<I: Iterator<Item = char>>(ch: char, char
 
 #[cfg(feature = "unicode-width")]
 #[inline]
-fn ch_width(ch: char) -> usize {
-    unicode_width::UnicodeWidthChar::width(ch).unwrap_or(0)
+fn ch_width(ch: char, tab_width: u8) -> usize {
+    if ch == '\t' {
+        tab_width as usize
+    } else {
+        unicode_width::UnicodeWidthChar::width(ch).unwrap_or(0)
+    }
 }
 
 /// First character which [`ch_width`] will classify as double-width.
@@ -72,8 +76,10 @@ const DOUBLE_WIDTH_CUTOFF: char = '\u{1100}';
 
 #[cfg(not(feature = "unicode-width"))]
 #[inline]
-fn ch_width(ch: char) -> usize {
-    if ch < DOUBLE_WIDTH_CUTOFF {
+fn ch_width(ch: char, tab_width: u8) -> usize {
+    if ch == '\t' {
+        tab_width as usize
+    } else if ch < DOUBLE_WIDTH_CUTOFF {
         1
     } else {
         2
@@ -88,8 +94,8 @@ fn ch_width(ch: char) -> usize {
 /// ```
 /// use textwrap::core::display_width;
 ///
-/// assert_eq!(display_width("Café Plain"), 10);
-/// assert_eq!(display_width("\u{1b}[31mCafé Rouge\u{1b}[0m"), 10);
+/// assert_eq!(display_width("Café Plain", 0), 10);
+/// assert_eq!(display_width("\u{1b}[31mCafé Rouge\u{1b}[0m", 0), 10);
 /// ```
 ///
 /// **Note:** When the `unicode-width` Cargo feature is disabled, the
@@ -107,11 +113,11 @@ fn ch_width(ch: char) -> usize {
 /// ```
 /// use textwrap::core::display_width;
 ///
-/// assert_eq!(display_width("Cafe Plain"), 10);
+/// assert_eq!(display_width("Cafe Plain", 0), 10);
 /// #[cfg(feature = "unicode-width")]
-/// assert_eq!(display_width("Cafe\u{301} Plain"), 10);
+/// assert_eq!(display_width("Cafe\u{301} Plain", 0), 10);
 /// #[cfg(not(feature = "unicode-width"))]
-/// assert_eq!(display_width("Cafe\u{301} Plain"), 11);
+/// assert_eq!(display_width("Cafe\u{301} Plain", 0), 11);
 /// ```
 ///
 /// ## Emojis and CJK Characters
@@ -123,8 +129,8 @@ fn ch_width(ch: char) -> usize {
 /// ```
 /// use textwrap::core::display_width;
 ///
-/// assert_eq!(display_width("😂😭🥺🤣✨😍🙏🥰😊🔥"), 20);
-/// assert_eq!(display_width("你好"), 4);  // “Nǐ hǎo” or “Hello” in Chinese
+/// assert_eq!(display_width("😂😭🥺🤣✨😍🙏🥰😊🔥", 0), 20);
+/// assert_eq!(display_width("你好", 0), 4);  // “Nǐ hǎo” or “Hello” in Chinese
 /// ```
 ///
 /// # Limitations
@@ -150,9 +156,9 @@ fn ch_width(ch: char) -> usize {
 ///
 /// assert_eq!("👨‍🦰".chars().collect::<Vec<char>>(), ['\u{1f468}', '\u{200d}', '\u{1f9b0}']);
 /// #[cfg(feature = "unicode-width")]
-/// assert_eq!(display_width("👨‍🦰"), 4);
+/// assert_eq!(display_width("👨‍🦰", 0), 4);
 /// #[cfg(not(feature = "unicode-width"))]
-/// assert_eq!(display_width("👨‍🦰"), 6);
+/// assert_eq!(display_width("👨‍🦰", 0), 6);
 /// ```
 ///
 /// This happens because the grapheme consists of three code points:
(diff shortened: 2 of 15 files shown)


src/columns.rs
```diff
@@ -77,14 +77,14 @@ where
...
-        .saturating_sub(display_width(middle_gap) * (columns - 1));
+        .saturating_sub(display_width(left_gap, options.tab_width))
+        .saturating_sub(display_width(right_gap, options.tab_width))
+        .saturating_sub(display_width(middle_gap, options.tab_width) * (columns - 1));
 
     let column_width = std::cmp::max(inner_width / columns, 1);
     options.width = column_width;
     let last_column_padding = " ".repeat(inner_width % column_width);
-    let wrapped_lines = wrap(text, options);
+    let wrapped_lines = wrap(text, options.clone());
```
mgeisler, line 87: I suppose you could save the `clone()` call if you copy out `options.tab_width` earlier in the function?

src/columns.rs
```diff
@@ -94,7 +94,9 @@ where
             match wrapped_lines.get(line_no + column_no * lines_per_column) {
                 Some(column_line) => {
                     line.push_str(column_line);
-                    line.push_str(&" ".repeat(column_width - display_width(column_line)));
+                    line.push_str(
+                        &" ".repeat(column_width - display_width(column_line, options.tab_width)),
```
mgeisler, line 98: A bit of a side comment: it just occurred to me that it might be nice to do a 

```rust
let display_width = |text| core::display_width(text, options.tab_width);
```

early on in the function. Then everything can stay the same. I'm always on the fence with such things: I don't like indirection and extra layers, but here it might be nice since it cuts down on the repetition.

src/columns.rs
```diff
@@ -190,4 +192,13 @@ mod tests {
     fn wrap_columns_panic_with_zero_columns() {
         wrap_columns("", 0, 10, "", "", "");
     }
+
+    #[test]
+    fn wrap_columns_with_tabs() {
+        let options = Options::new(23).tab_width(4);
+        assert_eq!(
+            wrap_columns("hello\tthis\tis\tlong\tyeah", 2, options, ".", ".", "."),
+            vec![".hello     .is\tlong.", ".this      .yeah      ."]
```
mgeisler, line 201: The `.`s here confused me :smile: Could you use `|` or something similar instead so that it doesn't look like it's part of the text?

src/core.rs
```diff
@@ -307,11 +318,9 @@ impl Fragment for Word<'_> {
         self.width as f64
     }
 
-    // We assume the whitespace consist of ' ' only. This allows us to
-    // compute the display width in constant time.
```
mgeisler, line 311: This might be important... Have you tried

```sh
$ cd  benchmarks
$ cargo criterion fill_optimal_fit_unicode
```

and checked the graph in `target/criterion/reports/String%20lengths/fill_optimal_fit_unicode/index.html`?

It should look something like this so that we know that performance scales linearly:

image

mgeisler: I think the change here might be just fine: most `Word` values will have a single space as their whitespace, so the performance will still be linear (at least with the simple benchmark I've been using).

Also, I think I'll finally rename the benchmark to something else than `String lengths` :smile:

src/options.rs
```diff
@@ -118,6 +122,26 @@ impl<'a> Options<'a> {
         }
     }
 
+    /// Change [`self.tab_width`]. The tab width is how wide
+    /// a tab character should be considered. By default, a tab
+    /// character to be consiered to have a width of 0.
```
mgeisler, line 127: I'm actually thinking that we should use a more traditional width of 8. If we go through the effort of adding support for this, people should see it work out of the box — if people discover that they have unexpected TAB characters in their strings, they can either filter them out or set the width back to zero.

src/options.rs
```diff
@@ -118,6 +122,26 @@ impl<'a> Options<'a> {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use textwrap::{Options, wrap};
+    ///
+    /// let options = Options::new(16).tab_width(4);
+    /// assert_eq!(wrap("This\tis\tan\texample.", options),
+    ///            vec!["This\tis\tan",
```
mgeisler, line 136: If I didn't know anything about the crate, I think this example would confuse me: it looks like it shows that we split text on `\t`. However, I believe what actually happens here is that

1. `tab_width` is taken into account — the new behavior, which is great! — and then
2. `break_words` makes us break the line on the `\t`

It's hard to tell because `"This\tis\tan"` happens to have 8 letters and 2 TABs, each of which is 4 columsn wide. So the string fits precisely in the width of 16 columns allowed.

We should figure out if we start counting TAB as a word separator. Until then, a better illustration of `tab_width` would be something like this:

```rust
assert_eq!(wrap("foo\tbar baz", Options::new(12).tab_width(0)), vec!["foo\tbar baz"]);
assert_eq!(wrap("foo\tbar baz", Options::new(12).tab_width(4)), vec!["foo\tbar", "baz"]);
```

src/word_splitters.rs
```diff
@@ -204,9 +205,10 @@ where
             if prev < word.word.len() || prev == 0 {
                 let w = Word {
                     word: &word.word[prev..],
-                    width: display_width(&word[prev..]),
-                    whitespace: word.whitespace,
-                    penalty: word.penalty,
+                    // see comment by display_width call above regarding
```
mgeisler, line 208: I think "above" is a bit far away here. Also, it looks like `tab_width` is readily available here so we could just use it?

src/columns.rs
mtoohey31: Good suggestion, I think that does make things cleaner. I've made the change in 08e4d4a15c47a9093ef8e2afcb0bf2b0528941a1.

src/columns.rs
mtoohey31: Also fixed by 08e4d4a15c47a9093ef8e2afcb0bf2b0528941a1.

src/columns.rs
mtoohey31: That's fair, `'.'` is a bit unexpected for a column divider. I've changed that in 324ba420b4aad16860ef74fbff6fd4265b82b91c.

src/word_splitters.rs
mtoohey31: You're right, I think I just forgot to update this call when I updated the one above. Fixed in a901fce7b0165ba2186aad3e5fdca02b96976949.

src/core.rs
mtoohey31: These are the results as of a901fce7b0165ba2186aad3e5fdca02b96976949:

chart

So it looks like things are ok. My chart looks a little different than yours though, but I think it's mainly due to my default font being different.

src/options.rs
mtoohey31: Fair point! It's also a very good thing you suggested this, because it revealed a bug I'd introduced in `wrap_single_line`. It looks like the condition that I've updated there in 046911e8530bc70cee13de1f3485f56a646bae62 makes the assumption that a string's display width is no longer than its byte length, which is no longer true since tabs only take one byte, but may be configured to take up much more width than that. Does the change I've made there seem like a reasonnable solution to you?

src/options.rs
mtoohey31: I'd be fine with that; it's just a default anyways, and changing it is simple. How do you feel about this given https://github.com/mgeisler/textwrap/pull/490#discussion_r1018629643 though? If we set it to anything greater than 1, we loose the fast in `wrap_single_line` by default, at least with my current fix.

src/columns.rs
mgeisler: Nice!

src/core.rs
mgeisler: Cool, that looks all good to me.

mgeisler:
The fuzz tests need some small fixes, something like this:

```patch
modified   fuzz/fuzz_targets/wrap_first_fit.rs
@@ -14,12 +14,12 @@ struct Word {
 #[rustfmt::skip]
 impl core::Fragment for Word {
     fn width(&self) -> f64 { self.width }
-    fn whitespace_width(&self) -> f64 { self.whitespace_width }
+    fn whitespace_width(&self, _: u8) -> f64 { self.whitespace_width }
     fn penalty_width(&self) -> f64 { self.penalty_width }
 }
 
 fuzz_target!(|input: (f64, Vec<Word>)| {
     let width = input.0;
     let words = input.1;
-    let _ = wrap_first_fit(&words, &[width]);
+    let _ = wrap_first_fit(&words, &[width], 0);
 });
modified   fuzz/fuzz_targets/wrap_optimal_fit.rs
@@ -35,7 +35,7 @@ struct Word {
 #[rustfmt::skip]
 impl core::Fragment for Word {
     fn width(&self) -> f64 { self.width }
-    fn whitespace_width(&self) -> f64 { self.whitespace_width }
+    fn whitespace_width(&self, _: u8) -> f64 { self.whitespace_width }
     fn penalty_width(&self) -> f64 { self.penalty_width }
 }
 
@@ -57,5 +57,5 @@ fuzz_target!(|input: (usize, Vec<Word>, Penalties)| {
         }
     }
 
-    let _ = wrap_optimal_fit(&words, &[width as f64], &penalties);
+    let _ = wrap_optimal_fit(&words, &[width as f64], 0, &penalties);
 });
modified   fuzz/fuzz_targets/wrap_optimal_fit_usize.rs
@@ -35,7 +35,7 @@ struct Word {
 #[rustfmt::skip]
 impl core::Fragment for Word {
     fn width(&self) -> f64 { self.width as f64 }
-    fn whitespace_width(&self) -> f64 { self.whitespace_width as f64 }
+    fn whitespace_width(&self, _: u8) -> f64 { self.whitespace_width as f64 }
     fn penalty_width(&self) -> f64 { self.penalty_width as f64 }
 }
 
@@ -45,5 +45,5 @@ fuzz_target!(|input: (usize, Vec<Word>, Penalties)| {
     let width = input.0;
     let words = input.1;
     let penalties = input.2.into();
-    let _ = wrap_optimal_fit(&words, &[width as f64], &penalties);
+    let _ = wrap_optimal_fit(&words, &[width as f64], 0, &penalties);
 });
```

With that, a problem will appear: running `cargo fuzz run wrap_fast_path` fails on input like

```
    let input = (String::from("foo \t"), 20);
```

You'll get

```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `["foo \t"]`,
 right: `["foo "]`', fuzz_targets/wrap_fast_path.rs:20:5
```

The problem is that we end up with two words instead of just one:

```rust
[src/textwrap/src/word_separators.rs:293] &opportunities = IntoIter(
    [
        (
            4,
            Allowed,
        ),
        (
            5,
            Mandatory,
        ),
    ],
)
[src/wrap.rs:231] &words = [
    Word {
        word: "foo",
        whitespace: " ",
        penalty: "",
        width: 3,
    },
    Word {
        word: "",
        whitespace: "\t",
        penalty: "",
        width: 0,
    },
]
```

The fast-path optimization assumes that every word matters, but here the second word should actually have been merged with the first word (with `whitespace: " \t",`). So when the slow path runs, it will add `"foo"` (`word`), then `" "` (`whitespace`) and finally `""` (`word`) but no `whitespace` from the second word since it's the final word of the line.

I think we might be able to fix this if we run `trim_end_matches(&[' ', '\t'])` before finding words. We don't output the trailing whitespace any longer, so it should be correct to do that. This seems to work from some light testing:

```patch
modified   src/core.rs
@@ -243,7 +243,7 @@ impl<'a> Word<'a> {
     /// All trailing whitespace is automatically taken to be the whitespace part
     /// of the word.
     pub fn from(word: &str) -> Word<'_> {
-        let trimmed = word.trim_end();
+        let trimmed = word.trim_end_matches(&[' ', '\t']);
         Word {
             word: trimmed,
             // trimmed shouldn't contain whitespace, so we don't need to pass
modified   src/wrap.rs
@@ -203,7 +203,7 @@ pub(crate) fn wrap_single_line<'a>(
         options.subsequent_indent
     };
     if line.len() < options.width && options.tab_width <= 1 && indent.is_empty() {
-        lines.push(Cow::from(line.trim_end_matches(' ')));
+        lines.push(Cow::from(line.trim_end_matches(&[' ', '\t'])));
     } else {
         wrap_single_line_slow_path(line, options, lines)
     }
@@ -217,6 +217,8 @@ pub(crate) fn wrap_single_line_slow_path<'a>(
     options: &Options<'_>,
     lines: &mut Vec<Cow<'a, str>>,
 ) {
+    let line = line.trim_end_matches(&[' ', '\t']);
+
     let initial_width = options
         .width
         .saturating_sub(display_width(options.initial_indent, options.tab_width));
```

src/options.rs
mtoohey31: Looks like `fill`'s fast path had the same issue. I fixed it the same way in e9d18ce8da6fa8379e6fc278ccbf56c35b61720b.

mtoohey31:
Fuzz tests should :crossed_fingers: be all fixed in 8b2a9f18089f01e8d6b329581ebce8c9d2e32129, and the wasm demo should be fixed in f1352c88fdc6986eb0cbb8675a41f14502987102. I'm just ignoring the `tab_width` parameter in the wasm demo, because `web_sys::CanvasRenderingContext2d::measure_text` should already register tab widths correctly, right?

mtoohey31:
> Hey @mtoohey31! Happy new year... :smile:

Haha, thanks :slightly_smiling_face: Same to you!

> Can I get you to squash your fixes into the commits which introduced the problems in the first place? That will make the commits here easier to review. It would also be great if you can rebase on top of the latest `master` so we can get fresh test run.

Sure, should be rebased now! Pretty much all of the extra commits were fixes for problems, so I've squashed everything into one commit. Hope that's ok; if there was another way you'd like me to split things up though, let me know.

mgeisler:
> Pretty much all of the extra commits were fixes for problems, so I've squashed everything into one commit. Hope that's ok; if there was another way you'd like me to split things up though, let me know.

That is precisely what I was hoping for — fixes to problems found in PR discussions are not useful for the long-term history, so they should ideally be squashed away before merging. Thank you!

src/options.rs
mgeisler: > It looks like the condition that I've updated there in [046911e](https://github.com/mgeisler/textwrap/commit/046911e8530bc70cee13de1f3485f56a646bae62) makes the assumption that a string's display width is no longer than its byte length, which is no longer true since tabs only take one byte, but may be configured to take up much more width than that. Does the change I've made there seem like a reasonnable solution to you?

Oh, that is a good catch! I had not thought of that at all :sweat_smile: I think the fix looks good! You bring up a good point about the interaction of the fast path and the tab with in your other comment, so let me comment there as well.

callum-oakley:
Hi! What's the status of this? I'd be interested in helping to get it over the finish line if there's any more work required (I'm keen to get https://github.com/helix-editor/helix/issues/3622 fixed).

Also thanks for the excellent crate. :)

mtoohey31:
@mgeisler I see there are some conflicts now, I would still like to get this fixed at some point though. If I rebase the changes, would you possibly have time to look at the merge request again? Or is there an issue with the general approach that you'd like to resolve before we move forward? I don't mind starting the PR from scratch with a new approach if that's what has to happen.

mtoohey31:
Closing in favour of #539.

PR #498, opened 2022-11-20
Add CIFuzz Github action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR. This is a service offered by OSS-Fuzz where Textwrap already runs. In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

I noticed the CI already builds and runs the fuzzers -- the benefit of CIFuzz it has a set of features that are useful e.g. only highlighting issues only if they are introduced by the given PR, using of OSS-Fuzz code corpus and more. Let me know what you think!

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'textwrap'
+        dry-run: false
+        language: rust
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'textwrap'
+        fuzz-seconds: 300
+        dry-run: false
+        language: rust
+    - name: Upload Crash
+      uses: actions/upload-artifact@v3
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts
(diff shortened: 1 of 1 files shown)


mgeisler:
Hey @DavidKorczynski, this sounds great! Do I understand correctly that this will run all of our fuzz targets? How do I enable new targets in the future?

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'textwrap'
+        dry-run: false
```
mgeisler, line 12: The page you linked says that `dry-run: false` is the default value — I would thus prefer to remove this line.

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,26 @@
...
+        language: rust
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'textwrap'
+        fuzz-seconds: 300
+        dry-run: false
```
mgeisler, line 19: Same about this `dry-run`.

.github/workflows/cifuzz.yml
DavidKorczynski: Done, thanks!

DavidKorczynski:
> How do I enable new targets in the future?

The action relies on the `build.sh` script here: https://github.com/google/oss-fuzz/blob/master/projects/textwrap/build.sh

You can modify this script whenever you want! Do a PR on oss-fuzz and we'll merge it in. Another option is to make https://github.com/google/oss-fuzz/blob/master/projects/textwrap/build.sh call a script inside of this repo (it clones this repo anyways from the Dockerfile https://github.com/google/oss-fuzz/blob/master/projects/textwrap/Dockerfile), and, thus you can easily extend which fuzzers should be build

mgeisler:
> You can modify this script whenever you want! Do a PR on oss-fuzz and we'll merge it in.

Nice, thanks for explaining! I think this looks nice, so let's merge it. We can then remove the "Required" flag for the old GitHub actions and thus clean this up a bit.

PR #514, opened 2023-06-20
Add unit test for uncovered regions

CXWorks:
Hi,

Thanks for your time & patience to review this PR.

We are researchers focusing on Rust unit tests. By examine the existing code, we found a unit test can be added to improve the repo's overall unit test coverage(this project is already been well tested).
We only manually changed our assert statements to make it consistent with existing unit tests and the newly covered code region is:
https://github.com/mgeisler/textwrap/blob/872f221fede7aa78e730bb7895074455269a431a/src/refill.rs#L77-L80
Thanks again for reviewing.

### DIFF
--- a/src/refill.rs
+++ b/src/refill.rs
@@ -224,10 +224,10 @@ mod tests {
     }
 
     #[test]
-    fn test_unfill_missing() {
-        let (text, options) = unfill("í\n*\n/");
-        assert_eq!(text, "í * /");
-        assert_eq!(options.width, 1);
+    fn test_unfill_consecutive_different_prefix() {
+        let (text, options) = unfill("foo\n*\n/");
+        assert_eq!(text, "foo * /");
+        assert_eq!(options.width, 3);
         assert_eq!(options.line_ending, LineEnding::LF);
     }
 
--- a/src/refill.rs
+++ b/src/refill.rs
@@ -224,6 +224,14 @@ mod tests {
     }
 
     #[test]
+    fn test_unfill_missing() {
+        let (text, options) = unfill("í\n*\n/");
+        assert_eq!(text, "í * /");
+        assert_eq!(options.width, 1);
+        assert_eq!(options.line_ending, LineEnding::LF);
+    }
+
+    #[test]
     fn unfill_trailing_newlines() {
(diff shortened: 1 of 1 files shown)


CXWorks:
Thanks for running the CI check, it seems one of them failed due to nightly toolchain issue, is there anything I can do?

src/refill.rs
```diff
@@ -223,6 +223,14 @@ mod tests {
         assert_eq!(options.line_ending, LineEnding::LF);
     }
 
+    #[test]
+    fn test_unfill_missing() {
+        let (text, options) = unfill("í\n*\n/");
```
mgeisler, line 228: Could you simplify the test a bit? The `í` could probably be a `foo` and the final `/` could probably be something else, right? Basically, take the fuzzer output and turn it into something a human would write :smile:

src/refill.rs
```diff
@@ -223,6 +223,14 @@ mod tests {
         assert_eq!(options.line_ending, LineEnding::LF);
     }
 
+    #[test]
+    fn test_unfill_missing() {
```
mgeisler, line 227: The word `missing` is confusing here. I think the test is covering the case where a bullet point list is _not_  actually seen as a list?

mgeisler:
> Thanks for running the CI check, it seems one of them failed due to nightly toolchain issue, is there anything I can do?

I restarted the jobs — you could have done the same by pushing a different commit to the branch.

The PR looks good, great job at finding a test case to improve the coverage like this. I left two small comments about making the test case more human-friendly.

src/refill.rs
CXWorks: Hhh, thanks for your comment, I will manually change it for this one and try to add this feature.

src/refill.rs
CXWorks: I think you are right and I will manually change the name, thanks.

mgeisler (approved):
Awesome, thanks so much!

src/refill.rs
mgeisler: Thanks!

CXWorks:
Thanks for your feedback & comments, it's really helpful.

PR #520, opened 2023-09-29
Clarify that WordSeparator is an enum rather than a trait

jelmer:
Clarify that WordSeparator is an enum rather than a trait


mgeisler:
Thanks for catching this!

PR #521, opened 2023-10-18
Add `Options::width`

9999years:
This is a convenience method to set the width of the `Options` struct. I found it useful to have builder-style access to this property, and it matches some of the other provided methods.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -118,6 +118,13 @@ impl<'a> Options<'a> {
         }
     }
 
+    /// Set [`self.width`] to the given value.
+    ///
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Thanks, makes sense!

mgeisler:
Hi @9999years, thanks for improving the crate!

Over the years, I've been torn between going all in on the builder pattern and letting people configure the options using public fields — and as a result, we now have both :smile: However, I now lean more towards using a builder pattern throughout. Would you be up for making a PR which hides the fields and turns this into a "proper" builder?

PR #522, opened 2023-10-18
Use struct init shorthand `..self` in `Options`

9999years:
Drive-by refactor to remove some boilerplate. No functional changes.


mgeisler (approved):
Thanks, looks good!

PR #523, opened 2023-10-18
Add `wrap`, `fill`, `refill` methods to `Options`

9999years:
These functions all take an `Options` as a parameter. I usually construct an `Options` and then use it for multiple calls, so it's more convenient for me to have access to these as methods.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -1,5 +1,7 @@
 //! Options for wrapping text.
 
+use std::borrow::Cow;
+
 use crate::{LineEnding, WordSeparator, WordSplitter, WrapAlgorithm};
 
 /// Holds configuration options for wrapping and filling text.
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @9999years, thanks for the PR.

This change would make the API similar to what I had [before version 0.13 from 2020](https://github.com/mgeisler/textwrap/blob/master/CHANGELOG.md#wrapper-is-replaced-with-options). So I can fully understand why this API seems natural — I started out with the same one :smile: 

My feeling back then is summed up a bit in #213, but overall I was concerned about two things:

- a one-struct-to-rule-them-all in the form of `Wrapper` (back then) or `Options` (now). I wanted to avoid an always-expanding struct with more and more methods.
- having more than one way to call `wrap`, `fill`, and friends.

Since 2020, the library has also gotten some simple column wrapping functionality — but you didn't add those rarely-used methods to the `Options` struct here :smile: I would prefer not having to decide that is important enough to be added as methods to `Options`.

Does this seem reasonable?

9999years:
Hmm. If we're talking about breaking changes in #521 already, here's my proposal:

- `Options` struct with methods for formatting functions (`opts.wrap("foo")`, `opts.fill("foo")`, etc.)
- Top-level formatting functions with default options (`textwrap::wrap("foo")`, `textwrap::fill("foo")`, etc.)

mgeisler:
> * Top-level formatting functions with default options (`textwrap::wrap("foo")`, `textwrap::fill("foo")`, etc.)

I don't think this will work: we need to know the correct wrapping width at the very least. The library could of course make an "opinionated" guess by using 80 columns or so, but I feel that won't be very popular with most people.

I will not add back the methods that were removed back in 2020 because of the problems I mentioned above. Overall, I hope the API of the crate can:

- support simple use cases using a very simple API: `textwrap::fill("foo bar", 80)`.
- support medium-complexity use cases: `textwrap::wrap(text, options)`.
- support complex use cases by implementing [`Fragment`](https://docs.rs/textwrap/latest/textwrap/core/trait.Fragment.html).

I feel the current API does this pretty well. Having to import both `Options` and `wrap` or `fill` seems okay to me.

Issue #525, opened 2023-11-07
`display_width` and hyperlinks


tertsdiepraam:
Hi! I'm working on uutils and collaborating with the eza people. We have use case for the `display_width` function that requires hyperlinks to be ignored. This does not seem to be currently supported by `textwrap`. This is understandable since it's a fairly niche use case but we'd still very much like to see this supported.

As far as I can tell the issue boils down to the fact that `textwrap` assumes that the ANSI sequences start with `ESC [`, but hyperlinks are part of Operating System Command block that start with `ESC ]` (which is info I'm also just pulling from wikipedia, can't say I'm an expert at ANSI codes: https://en.wikipedia.org/wiki/ANSI_escape_code#OSC). After `ESC ]`, it looks like there are more characters allowed before the sequence ends and it must therefore end with `ESC \`.

If this is something you're interested in to support, I'd be happy to try to make a PR.

cc @pthorpe92

mgeisler:
Hi! Yes, this would be a great addition and I would happy to support it.

I'm also not an expert on this, but I think you read the standard correctly. I found https://github.com/Alhadis/OSC8-Adoption which has lots of good links to other projects supporting this.

Please put up a PR and I'll take a look!

tertsdiepraam:
Great! I've also been putting an implementation together in a separate repo: <https://github.com/uutils/ansi-width/>, which we intend to publish as a crate. Would you accept using that library in `textwrap`?[^1] We are going to use that in some of our other libraries, so it will be maintained. If you don't want that, I completely understand, I'd still be happy to just implement a PR for `textwrap`.

[^1]: I don't have the choice between simple width and unicode width in there yet, but that could be easily added.

mgeisler:
I've gotten pushback in the past for any dependency on Textwrap — this is actually why all the dependencies are optional now since people seem to care more about compilation times than they do about features. Despite that, Clap decided to vendor Textwrap inside their source tree: [`clap_builder/src/output/textwrap`](https://github.com/clap-rs/clap/tree/master/clap_builder/src/output/textwrap). That crate was a big reverse dependency in the past.

So I'm in doubt if I should take on a dependency for this. People have suggested this in the past, but with larger crates which implements support for parsing more or all of the ANSI escape codes.

[...]

PR #526, opened 2023-11-13
`display_width`: ignore ANSI hyperlinks in width calculation

tertsdiepraam:
Closes https://github.com/mgeisler/textwrap/issues/525

I've attempted to keep the implementation similar to the current implementation and added a simple test for it. This keeps it in a bit of an awkward state (e.g. comparing with `CSI.1` in one branch and `']'` in the other). Happy to take suggestions for the direction to take this PR!

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -60,10 +60,11 @@ pub(crate) fn skip_ansi_escape_sequence<I: Iterator<Item = char>>(ch: char, char
         } else if next == Some(']') {
             // We have found the start of an Operating System Command, which
             // extends until the next sequence "\x1b\\" (the String Terminator
-            // sequence).
+            // sequence) or the BEL character. The BEL character is non-standard,
+            // but it is still used quite often, for example, by GNU ls.
             let mut last = ']';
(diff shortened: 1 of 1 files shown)


anthraxx:
I was actually hit by this bug and found this PR. @mgeisler any chance you can take a look :cat:

tertsdiepraam:
I added support for a BEL character as separator between the link and the text. This is used by some programs such as GNU `ls` and therefore also uutils `ls`. Since this PR has not seen any activity, I'll be moving uutils to my own crate, but I figured I would still fix the same issue here. I think it would still be nice for the community to have this PR merged!

mgeisler (approved):
Thanks a lot for this!

mgeisler:
Thanks for the ping! I'm sorry I had forgotten about this... let's merge it right away so I can make a new release.

tertsdiepraam:
Thanks!

mgeisler:
I finally got around to merge the release PR: #533 and this means 0.16.1 is available!

PR #540, opened 2024-06-26
core: avoid spurious f64 cast and comparison

lucab:
This tweaks `break_words()` logic by dropping a cast to f64 and a float comparison, directly using the integer width value instead.


mgeisler:
Thanks, looks great!

Let me fix the broken test so I can merge this.

PR #559, opened 2024-09-25
Reduce typo count.

waywardmonkeys:

PR #560, opened 2024-09-26
Update MSRV to 1.66 because of `unicode-width` dep

waywardmonkeys:
The `unicode-width` crate needs 1.66 because of using the `wrapping_add_signed` function.


waywardmonkeys:
This should fix the CI failure seen in #557 and #559.

If this lands (or CI is fixed in some other way), I have some other small PRs to submit.

mgeisler:
Thank you, this looks lovely!

PR #562, opened 2024-09-29
chore(deps): Update `terminal_size`, `unicode-width`, `termion`

waywardmonkeys:
* `terminal_size` is now using 0.4.0
* `termion` is now using 4.0.2
* `unicode-width` is now using 0.2.0


mgeisler:
Thanks! The major version bumps are the most important in my eyes — bumping SemVer-compatible versions is not strictly necessary, but it's fine since you're already touching the `Cargo.toml` file.

PR #563, opened 2024-09-29
chore: cargo fmt

waywardmonkeys:


waywardmonkeys:
I guess several other things need fixing for the new CI check to pass ... what do you think about if it should be done or not?

mgeisler:
> I guess several other things need fixing for the new CI check to pass ... what do you think about if it should be done or not?

I would prefer to not enforcing this in CI — the library doesn't get a lot of contributions, but when it does get a patch, there is no need to be super demanding on this. I don't feel this has been a problem in the past.

I like `dprint` and use it in other projects, so feel free to update the PR to keep the config file.

mgeisler:
Let's get this merged so I can make a new shiny release!

PR #564, opened 2024-09-29
chore(ci): Update to `actions/checkout@v4`

waywardmonkeys:


mgeisler:
Thanks! This reminded me that I should setup Dependabot, so this is now done via #567.

PR #579, opened 2025-03-03
Bump `rust-version` to 1.70

mgeisler:
This does two things: it fixes the warnings about the derived `Debug` implementation after #557. It also means that I don’t have to deal with the slow Git-based crates.io registry checkout when testing the old version.


notgull:
@mgeisler Please consider lowering the MSRV back to at least v1.63, as this means we can no longer build this crate on Debian Stable.

ref: https://github.com/Smithay/calloop/pull/229

Issue #503, opened 2023-02-10
Option to preserve trailing whitespace


heiskane:
Currently working on something that requires me to keep track of a cursor in a message and wrap the message in a textbox with the cursor in the right place. This currently is not possible as wrapping trims the trailing whitespace making wrapped text different length from the original message misaligning any position i might track in the message.  

Trimming trailing whitespace could be disabled in the options?

wanted behavior.
```rust
let text = "as     a hello";
let wrapped = textwrap::wrap(text, 5);
// ["as   ", "  a ", "hello"]
```

mgeisler:
Hi @heiskane,

> Currently working on something that requires me to keep track of a cursor in a message and wrap the message in a textbox with the cursor in the right place.

Interesting! Since you're doing something custom, have you considered using the [`Word`](https://docs.rs/textwrap/latest/textwrap/core/struct.Word.html) struct directly? You might even want to implement [`Fragment`](https://docs.rs/textwrap/latest/textwrap/core/trait.Fragment.html) for your own custom type.

All the wrapping logic operates on fragments: the "upper layer" simply splits the text into fragments using different options.

I think your use case could be handled by

- creating a `Fragment` for every space character and
- moving the whitespace to the "word" part of the `Fragment` instead of the "whitespace" part

Today, the string `"as     a hello"` is turned into words and whitespace like this:

```
[("as", "     "), ("a", " "), ("hello". "")]
```

I'm suggesting instead turning it into

```
[("as", ""), (" ", ""), (" ", ""), (" ", ""), (" ", ""), (" ", ""), ("a", ""), (" ", ""), ("hello", "")]
```

That should do the trick. You should be able to do this with the [`WordSeparator::Custom`](https://docs.rs/textwrap/latest/textwrap/enum.WordSeparator.html) setting.

heiskane:
Thanks for the reply. I experimented a bit with what you suggested but i am still having some issues.  

The issue with this is that [WordSeparator::Custom](https://docs.rs/textwrap/latest/textwrap/enum.WordSeparator.html) takes a closure that returns iterator over `Word`s and `Word::from()` strips whitespace so you can't really do something like `Word::from(" ")`. I can't construct `Word` manually either because it needs `width` which is a private field. Not sure how i can make this happen with these restrictions.  

[...]

PR #586, opened 2025-04-03
Make `Word::width` public so it can be used without Word::from

nstolpe:


mgeisler:
Hi @nstolpe, sorry about the slow response! I've been too busy to look after this repository :cry: I think the use case in #503 is good, so let's merge this!

PR #587, opened 2025-04-18
Add `Options::preserve_trailing_space`

HidenoriKobayashi:
Add a new option to preserve trailing spaces at the end of the line.

The default is false, which preserves the current behavior of trimming trailing spaces at the end. If set to true, it will preserve them.


HidenoriKobayashi:
As discussed offline, I'm creating this PR to run more tests on the change.

The documentation comments are just proposals. Please feel free to change it to suit your taste.  
Also if you think we need more tests, I'm happy to add them.

mgeisler (approved):
Thank you, this looks great!

PR #592, opened 2025-09-03
`Options`: Replace `&str` with `Cow`

9999years:
In [`tracing-human-layer`](https://docs.rs/tracing-human-layer/latest/tracing_human_layer/index.html), I would like users to be able to specify `textwrap::Options` ahead of time, but this is difficult because I can't hold an owned `Options` struct.

If we use a `Cow<'a, str>` instead of an `&'a str`, we'll be able to adapt existing use-cases while making it possible to construct a `textwrap::Options<'static>` with owned `String`s.

This is a breaking change.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -1,5 +1,7 @@
 //! Options for wrapping text.
 
+use std::borrow::Cow;
+
 use crate::{LineEnding, WordSeparator, WordSplitter, WrapAlgorithm};
 
 /// Holds configuration options for wrapping and filling text.
@@ -12,10 +14,10 @@ pub struct Options<'a> {
     pub line_ending: LineEnding,
     /// Indentation used for the first line of output. See the
     /// [`Options::initial_indent`] method.
-    pub initial_indent: &'a str,
+    pub initial_indent: Cow<'a, str>,
     /// Indentation used for subsequent lines of output. See the
     /// [`Options::subsequent_indent`] method.
-    pub subsequent_indent: &'a str,
+    pub subsequent_indent: Cow<'a, str>,
     /// Allow long words to be broken if they cannot fit on a line.
     /// When set to `false`, some lines may be longer than
     /// `self.width`. See the [`Options::break_words`] method.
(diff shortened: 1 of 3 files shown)


9999years:
Yeah, I currently have a `TextWrapOptionsOwned` struct which just has a copy of each of the fields of `textwrap::Options`, but owned:

https://docs.rs/tracing-human-layer/0.2.1/tracing_human_layer/struct.TextWrapOptionsOwned.html

9999years:
But I do like the idea of users being able to implement their own options with `&'static str`s and not have to heap-allocate them, which pushed me towards the `Cow` solution.

mgeisler:
> But I do like the idea of users being able to implement their own options with `&'static str`s and not have to heap-allocate them, which pushed me towards the `Cow` solution.

Yeah, I just don't like having to force people to go from `initial_indent("* ")` to `initial_indent(Cow::from("* "))` or similar. The conversion is unnecessary in most cases.

For when it is necessary because you have dynamic heap-allocated strings, well then a local owned type can still be used.

Comparing [`Options`](https://docs.rs/textwrap/latest/src/textwrap/options.rs.html#8-33) and [`TextWrapOptionsOwned`](https://docs.rs/tracing-human-layer/0.2.1/src/tracing_human_layer/textwrap.rs.html#31-48), I wonder why you need the owned type?

That is, I see that you don't use the `initial_indent` or `subsequent_indent` fields. So is your `TextWrapOptionsOwned` not nearly identical to `Options<'static>`?

9999years:
> Yeah, I just don't like having to force people to go from `initial_indent("* ")` to `initial_indent(Cow::from("* "))` or similar. The conversion is unnecessary in most cases.

Agreed. The new function signatures just require a `From` bound, so most invocations should still compile:

```rust
pub fn initial_indent<T>(self, initial_indent: T) -> Self
where
    Cow<'a, str>: From<T>,
{
    Options {
        initial_indent: initial_indent.into(),
        ..self
    }
}
```

> That is, I see that you don't use the `initial_indent` or `subsequent_indent` fields. So is your `TextWrapOptionsOwned` not nearly identical to `Options<'static>`?

Ah, that's true! Sorry, it's been a bit since I wrote this and the code has evolved a bit. I _mostly_ want the textwrap options to stay the same across the program, so I store a `TextWrapOptionsOwned`, but then I want to vary the `initial_indent` and `subsequent_indent` depending on the log level of the message being processed, so those get stored in a separate [`Style` struct](https://docs.rs/tracing-human-layer/latest/tracing_human_layer/struct.Style.html).

Your call, I could go either way on this.

mgeisler:
> Agreed. The new function signatures just require a `From` bound, so most invocations should still compile:

Right, but the internal representation becomes more complex than it needs to be. I'll let `Options` stay how it is right now, and then I hope this issue and the newly added example (thanks for the inspiration!) in #594 is enough to show people how they can make Textwrap work for their particular needs.

PR #608, opened 2026-04-16
word_separators: switch from unicode-linebreak to icu_segmenter

### LINKED ISSUE #334: Provide a WordSeparator which uses rust_icu
With #332 merged, it is now possible to customize the way text is split into words via the  `WordSeparator` trait. From the discussion in #220, I understand that the [rust_icu crate](https://docs.rs/rust_icu) is the gold standard for breaking text into words. We should add a `WordSeparator` implementation which uses that library.

durin42:
The former hasn't been touched in a while and has a dep on an old hashbrown, which is what caused me to notice the dep in the first place. icu_segmenter is maintained by the Unicode Consortium which seems like a more canonical solution anyway.

Closes #334.


mgeisler:
Hi @durin42, thanks a lot, it looks great! It even closes #334 :tada:

mgeisler:
I've been meaning to move towards the ICU crates, so I'm happy to merge this. One detail, though, is the binary size which is now 300 kB larger:

Before with `unicode-linebreak`:

| Configuration                            |  Binary Size |    Delta |
| :---                                     |         ---: |     ---: |
| quick-and-dirty implementation           |       320 KB |     — KB |
| textwrap without default features        |       339 KB |    18 KB |
| textwrap with smawk                      |       352 KB |    32 KB |
| textwrap with unicode-width              |       348 KB |    28 KB |
| textwrap with unicode-linebreak          |       365 KB |    45 KB |

After with `icu_segmenter`:

| Configuration                            |  Binary Size |    Delta |
| :---                                     |         ---: |     ---: |
| quick-and-dirty implementation           |       320 KB |     — KB |
| textwrap without default features        |       339 KB |    18 KB |
| textwrap with smawk                      |       352 KB |    32 KB |
| textwrap with unicode-width              |       348 KB |    28 KB |
| textwrap with unicode-linebreak          |       727 KB |   407 KB |

(This is from `cargo run --example make-table` in `examples/binary-sizes`.)

If this is a concern for someone, I believe we can tweak the API further by exposign the ICU [`DataProvider`](https://docs.rs/icu_provider/2.2.0/icu_provider/trait.DataProvider.html).

Cargo.toml
```diff
@@ -28,15 +28,16 @@ all-features = true
...
 [features]
 default = ["unicode-linebreak", "unicode-width", "smawk"]
+unicode-linebreak = ["icu_segmenter"]
 
 [lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ["cfg(fuzzing)"] }
 
 [dependencies]
 hyphenation = { version = "0.8.4", optional = true, features = ["embed_en-us"] }
+icu_segmenter = { version = "1.5", optional = true }
```
sffc, line 38: 1.5 is from a year and a half ago and won't be updated any more; you should migrate to 2.2

src/word_separators.rs
```diff
@@ -18,6 +18,11 @@
 use crate::core::skip_ansi_escape_sequence;
 use crate::core::Word;
 
+#[cfg(feature = "unicode-linebreak")]
+thread_local! {
+    static LINE_SEGMENTER: icu_segmenter::LineSegmenter = icu_segmenter::LineSegmenter::new_auto();
```
sffc, line 23: `new_auto()` is the correct default API. As @mgeisler noted, it carries a bit of binary size, mostly to support Southeast Asian languages that need an LSTM model, but it is substantially less than the binary size of ICU4C's line segmenter. We have ongoing work to further reduce the binary size, which you will pick up automatically if you use a modern version of icu4x.

mgeisler: Hi @sffc, awesome thanks a lot! I'll make sure to update the dependency and look forward to future updates.

PR #610, opened 2026-04-29
Bump icu_segmenter dependency to 2.3.0

mgeisler:
The 2.x release of `icu_segmenter` introduces support for complex scripts without word spaces (such as Thai, Lao, Khmer, and Burmese) using dictionary and LSTM neural network models via the `auto` feature. Enabling these models increases stripped binary size by
approximately 336 KB:

Textwrap has never had support for these scripts, neither with the old `unicode-linebreak` crate, nor with `icu_segmenter` version 1.5. To preserve a small default binary size, complex script segmentation is put behind a new opt-in feature, `unicode-linebreak-complex`. The binary sizes now look like this:

| Configuration                            |  Binary Size |    Delta |
| :---                                     |         ---: |     ---: |
| quick-and-dirty implementation           |       309 KB |     — KB |
| textwrap without default features        |       329 KB |    19 KB |
| textwrap with smawk                      |       342 KB |    32 KB |
| textwrap with unicode-width              |       338 KB |    28 KB |
| textwrap with unicode-linebreak          |       383 KB |    73 KB |
| textwrap with unicode-linebreak-complex  |       718 KB |   408 KB |

The minimum supported Rust version (`rust-version`) is bumped to 1.90 as required by `icu_segmenter` 2.3.0.


xtqqczze:
`icu_segmenter` 2.3.0 has a MSRV of 1.88, we should reduce MSRV to that if possible.

mgeisler:
> `icu_segmenter` 2.3.0 has a MSRV of 1.88, we should reduce MSRV to that if possible.

Thanks for #628, it'll merge in a minute.

Just out of curiosity, is there anything that restricts the MSRV for you? A distro shipping only Rust 1.88 or so?

xtqqczze:
> > `icu_segmenter` 2.3.0 has a MSRV of 1.88, we should reduce MSRV to that if possible.
> 
> Thanks for #628, it'll merge in a minute.
> 
> Just out of curiosity, is there anything that restricts the MSRV for you? A distro shipping only Rust 1.88 or so?

For `uutils`, Rust 1.88 was used because it was the latest version available in Ubuntu at the time. We haven't had a good reason to bump further since then. Would you cut a new `textwrap` release to unblock https://github.com/uutils/coreutils/pull/14486?

PR #617, opened 2026-06-14
Simplify by using `borrow()` instead of `borrow_mut()`

kokorokonekuto:
since we're just reading the value and not modifying it, we don't need it to be mutable.


mgeisler:
Nice, thanks for catching this!

PR #628, opened 2026-09-12
Reduce MSRV to 1.88

xtqqczze:
MSRV was bumped to 1.90 in https://github.com/mgeisler/textwrap/pull/610, with the justification that this was required by `icu_segmenter` 2.3.0. However, `icu_segmenter` 2.3.0 actually only requires Rust 1.88, so the MSRV can be reduced accordingly.


mgeisler:
Ah, cool, happy to merge this if tests pass!

PR #630, opened 2026-09-13
fix: manual_readme cargo lint

xtqqczze:


mgeisler:
Thanks, I saw warnings about this so it's also implied now?

xtqqczze:
https://doc.rust-lang.org/beta/cargo/reference/lints.html#manual_readme

PR #632, opened 2026-09-13
fuzz: fix lints

xtqqczze:


mgeisler:
Nice cleanup!

xtqqczze:
You could consider checking lints for `fuzz` in CI:

```sh
cd fuzz
cargo +nightly fmt --check
cargo clippy --workspace --all-targets --all-features -- -D warnings
```

but that might be overkill?

PR #647, opened 2026-09-17
Avoid arithmetic overflow in the column padding

youdie006:
Avoid arithmetic overflow in the column padding

`wrap_columns` subtracts the line width from the column width to size the
padding. When a wrapped line is wider than its column that subtraction
underflows, so the call panics in debug with "attempt to subtract with
overflow" and in release with "capacity overflow" once the wrapped value
reaches `" ".repeat()`. The documented panic is only for `columns` being
zero.

Two paths reach it with the public API. Disabling `break_words` and
passing a word wider than a column is the case the function's own
documentation describes, where it recommends `break_words` to prevent
words from protruding into the margins. The other needs no configuration
at all, because a double-width grapheme cannot be broken down to fit a
column of width 1.

Saturate the subtraction, which is what the other three width
subtractions in this function already do and what #623 did for the
multiplication last month. Every input that works today is unchanged.
Only the panicking ones differ, and they get zero padding for a line
that protrudes, exactly as the documentation describes.

This change was written with AI assistance (Claude). The reasoning and the
measurements were reproduced and checked by me before opening this pull
request.


mgeisler:
Thanks, this looks great!

